PR #272, opened 2025-04-16
feat(trace): add otel tracing

ionmincu:
I incorporated an LlmOpsHttpExporter that's compatible with OpenTelemetry, and a @traced() attribute that uses OpenTelemetry. Also, I've added a TracedDecoratorRegistry that I'll use in UiPath-langchain to replace the Langchain tracer instead of OpenTelemetry. Consequently, all @traced() methods will behave differently.

Add open telemetry tracing

Example

```python
from time import sleep
from opentelemetry import trace
import os

from aio_tracer import traced, wait_for_tracer
import asyncio

tracer = trace.get_tracer(__name__)

@traced()
def print_env_vars():
    for key, value in os.environ.items():
        print(f"{key}={value}")


@traced()
def sync_function(x, y):
    sleep(1)
    return x + y

@traced()
async def async_function(x, y):
    await asyncio.sleep(1)
    return x * y

@traced()
def generator_function(n):
    for i in range(n):
        yield i ** 2

@traced()
async def async_generator_function(n):
    for i in range(n):
        await asyncio.sleep(0.5)
        yield i ** 3

def main():
    main_inner()
    wait_for_tracer()
    
@traced()
def main_inner():
    print_env_vars()

    print("Sync function result:", sync_function(3, 4))

    with tracer.start_as_current_span("foo") as span:
        with tracer.start_as_current_span("bar") as span2:
            print("Hello world!")
            for value in generator_function(5):
                print("Generator yielded:", value)
    

@traced()
async def async_demo():
    print("Async function result:", await async_function(2, 5))
    async for value in async_generator_function(3):
        print("Async generator yielded:", value)

if __name__ == "__main__":
    main()

```

### DIFF
--- a/src/uipath/tracing/_traced.py
+++ b/src/uipath/tracing/_traced.py
@@ -1,0 +1,165 @@
+import inspect
+import json
+import logging
+from functools import wraps
+from typing import Any
+
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+from ._otel_exporters import LlmOpsHttpExporter
+from ._utils import _SpanUtils
+
+logger = logging.getLogger(__name__)
+
+trace.set_tracer_provider(TracerProvider())
+trace.get_tracer_provider().add_span_processor(BatchSpanProcessor(LlmOpsHttpExporter()))  # type: ignore
+tracer = trace.get_tracer(__name__)
+
+
+def wait_for_tracers():
+    """Wait for all tracers to finish."""
+    trace.get_tracer_provider().shutdown()  # type: ignore
+
+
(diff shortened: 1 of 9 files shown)


src/uipath/tracing/_traced.py
```diff
@@ -0,0 +1,165 @@
...
+                    "inputs",
+                    SpanUtils.format_args_for_trace_json(
+                        inspect.signature(func), *args, **kwargs
+                    ),
+                )
+                try:
+                    result = func(*args, **kwargs)
+                    span.set_attribute(
+                        "output", json.dumps(result, default=str)
+                    )  # Record output
```
gheorghitahurmuz, line 61: useless comment

gheorghitahurmuz (approved):

PR #307, opened 2025-04-29
chore: add input file option

radu-mocanu:
- input file option
- log success message after debug run

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-nightly==2.0.33.dev1003070120",
  
  # Any version from PR
  "uipath-nightly>=2.0.33.dev1003070000,<2.0.33.dev1003080000"
]
```

### DIFF
--- a/src/uipath/_cli/cli_run.py
+++ b/src/uipath/_cli/cli_run.py
@@ -1,6 +1,5 @@
 # type: ignore
 import asyncio
-import logging
 import os
 import traceback
 from os import environ as env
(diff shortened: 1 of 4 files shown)


cristipufu (approved):

src/uipath/_cli/cli_run.py
```diff
@@ -91,7 +90,7 @@ async def execute():
         )
     except Exception as e:
         # Handle unexpected errors
-        logger.exception("Unexpected error in Python runtime middleware")
+        console.info("Unexpected error in Python runtime middleware")
```
radu-mocanu, line 93: Note: we are logging the message here as 'info' to avoid exiting (and return failure Middleware result)

PR #336, opened 2025-05-08
feat: telemetry

caleniuc:
### Description

This introduces a telemetry decorator that captures and logs function invocation data.

#### Logs
- function name
- invocation timestamp
- tenant id
- organization id
- url
- sdk version
- python version
- os

### Usage
Use with the `@track` decorator:
- `@track`
- `@track()`
- `@track('name')`
- `@track(extra={attributeKey: 'attributeValue'})`
- `@track('name', extra={attributeKey: 'attributeValue'})`

Control via the `TELEMETRY_ENABLED` env var - `true/false`. (enabled by default)

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-nightly==2.0.52.dev1003360261",

  # Any version from PR
  "uipath-nightly>=2.0.52.dev1003360000,<2.0.52.dev1003370000"
]
```


.github/workflows/build.yml
```diff
@@ -26,6 +26,17 @@ jobs:
...
+      - name: Replace connection string placeholder
+        run: |
+          originalfile="src/uipath/telemetry/_constants.py"
+          tmpfile=$(mktemp)
+          trap 'rm -f "$tmpfile"' EXIT
+
+          rsync -a --no-whole-file --ignore-existing "$originalfile" "$tmpfile"
+          envsubst '$CONNECTION_STRING' < "$originalfile" > "$tmpfile" && mv "$tmpfile" "$originalfile"
+        env:
+          CONNECTION_STRING: ${{ secrets.APPINS_CONNECTION_STRING }}
```
cristipufu, line 38: Please discuss if we should have an API that returns the connection string (eg: if we decide to change it)

src/uipath/_cli/cli_run.py
```diff
@@ -108,6 +109,7 @@ async def execute():
     type=click.Path(exists=True),
     help="File path for the .json input",
 )
+@track
```
cristipufu, line 112: We should exclude runtime runs (eg env having a job id)

.github/workflows/build.yml
cristipufu: Also please make sure that having an invalid key fails silently and doesn't block execution

src/uipath/_cli/cli_run.py
ionmincu: See list of full envs from a run:

https://uipath.atlassian.net/wiki/spaces/Agents/pages/88737415226/Environment+Variables

gheorghitahurmuz (approved):

PR #356, opened 2025-05-12
CLI improvements

radu-mocanu:
- add --force flag for auth
- retrieve process details by package name and version (in case it gets renamed in pw)

### DIFF
--- a/src/uipath/_cli/cli_invoke.py
+++ b/src/uipath/_cli/cli_invoke.py
@@ -24,7 +24,7 @@ console = ConsoleLogger()
 client = httpx.Client(follow_redirects=True, timeout=30.0)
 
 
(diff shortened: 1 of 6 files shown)


src/uipath/_cli/cli_invoke.py
```diff
@@ -24,7 +24,7 @@
 client = httpx.Client(follow_redirects=True, timeout=30.0)
 
 
-def _read_project_name() -> str:
+def _read_project_details() -> Tuple[str, str]:
```
gheorghitahurmuz, line 27: also here: https://docs.python.org/3/library/typing.html#annotating-tuples

gheorghitahurmuz (approved):

PR #379, opened 2025-05-16
chore: update GitHub workflows for improved CI/CD process

gheorghitahurmuz:


.github/workflows/build.yml
```diff
@@ -10,21 +10,29 @@ on:
...
+        uses: actions/setup-python@v5
         with:
           python-version-file: ".python-version"
 
-      - name: Setup venv
-        run: |
-          uv venv
-          uv sync --all-extras
+      - name: Install dependencies
+        run: uv sync --all-extras
```
radu-mocanu, line 35: should we maybe also use `--refresh` here? https://docs.astral.sh/uv/reference/cli/#uv-run--refresh

.github/workflows/ci.yml
```diff
@@ -4,16 +4,15 @@ on:
...
+    paths-ignore:
+      - pyproject.toml
+  pull_request:
     branches:
       - main
 
 jobs:
   commit-lint:
-    if: ${{ github.event_name == 'pull_request_target' }}
+    if: ${{ github.event_name == 'pull_request' }}
```
radu-mocanu, line 15: isn't this unsafe?

PR #412, opened 2025-06-26
Fix connection tokenType query param

edis-uipath:
Fix:

The query string param for the connection token endpoint should be 'tokenType=direct' so we get a vendor token instead of an uipath token.


edis-uipath:
https://github.com/UiPath/uipath-python/pull/490

PR #512, opened 2025-08-26
fix: configurable scope for client credentials auth

cristipufu:


src/uipath/_cli/cli_auth.py
```diff
@@ -81,18 +81,25 @@ def set_port():
     required=False,
     help="Base URL for the UiPath tenant instance (required for client credentials)",
 )
+@click.option(
+    "--scope",
+    required=False,
```
razvalex, line 86: it works to set required=True if it has a default configured

ionmincu (approved):

PR #541, opened 2025-09-05
fix: check for spaces in project name

PopescuTudor:


src/uipath/_cli/_utils/_project_files.py
```diff
@@ -103,6 +103,13 @@ def validate_config(config: dict[str, str]) -> None:
             "'requires-python' field cannot be empty. Please specify it in pyproject.toml:  requires-python = \">=3.10\""
         )
 
+    if " " in config["project_name"]:
```
radu-mocanu, line 106: I think we should search `invalid_chars` beofore spaces. 
https://github.com/UiPath/uipath-python/pull/541/files#diff-630821237d5dd9028189bc40e2be3b9534ac83a3d4f2f3aee167b50b0ef70b84R113

PR #557, opened 2025-09-09
Added more verbose message for authentication failures

uipreliga:


src/uipath/models/errors.py
```diff
@@ -10,7 +10,7 @@ def __init__(
 class SecretMissingError(Exception):
     def __init__(
         self,
-        message="Authentication required. Please run \033[1muipath auth\033[22m.",
+        message="Authentication required. Please run \033[1muipath auth\033[22m or provide a valid secret in UIPATH_ACCESS_TOKEN (or UNATTENDED_USER_ACCESS_TOKEN) environmental variable.",
```
radu-mocanu, line 13: `UNATTENDED_USER_ACCESS_TOKEN` should not be configured by the user. 
I am not sure we want to have those error messages. In theory the CLI should take care of authentication. If the user wants to use some custom base_url/secret for some reason they will most likely do something like this:
```
from uipath import UiPath
uipath = UiPath(
    base_url="https://example.com",
    secret="my-secret-value"
)
```
 Maybe we can have a dedicated docs section for this and reference it in the error messages. @cristipufu what do you think?

uipreliga: Removed UNATTENDED_USER_ACCESS_TOKEN in the latest push. For me the initial error message wasn't very clear - hope that this one will reduce confusion among users.

src/uipath/models/errors.py
```diff
@@ -1,7 +1,7 @@
 class BaseUrlMissingError(Exception):
     def __init__(
         self,
-        message="Authentication required. Please run \033[1muipath auth\033[22m.",
+        message="Authentication required. Please run \033[1muipath auth\033[22m or provide a valid base URL in UIPATH_URL environmental variable.",
```
cristipufu, line 4: I think the correct term is "environment variable"

cristipufu (commented):
Let's also squash the commits please

PR #558, opened 2025-09-09
Allow retrieval of job status

akshaylive:
This change allows SDK consumers to retrieve job details of a started job.


cristipufu:
We already have it here: https://github.com/UiPath/uipath-python/blob/main/src/uipath/_services/jobs_service.py

Issue #530, opened 2025-09-04
PydanticDeprecatedSince20: `json_encoders` is deprecated.


cristipufu:
In Pydantic v1, you could define `class Config: json_encoders = {...}` to control serialization.

In v2, that’s replaced with `field_serializer` or `@model_serializer`.

PR #560, opened 2025-09-10
fix(json_encoder): adapt json serialization to pydantic v2

giuliastf:
Fix: https://github.com/UiPath/uipath-python/issues/530

https://docs.pydantic.dev/latest/concepts/serialization/#custom-serializers

### DIFF
--- a/src/uipath/models/action_schema.py
+++ b/src/uipath/models/action_schema.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from typing import List, Optional
 
 from pydantic import BaseModel, ConfigDict, Field
@@ -16,7 +15,6 @@ class ActionSchema(BaseModel):
         use_enum_values=True,
         arbitrary_types_allowed=True,
         extra="allow",
-        json_encoders={datetime: lambda v: v.isoformat() if v else None},
     )
 
     key: str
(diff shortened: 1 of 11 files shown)


src/uipath/models/action_schema.py
```diff
@@ -16,9 +16,14 @@ class ActionSchema(BaseModel):
         use_enum_values=True,
         arbitrary_types_allowed=True,
         extra="allow",
-        json_encoders={datetime: lambda v: v.isoformat() if v else None},
     )
 
+    @field_serializer("*", when_used="json")
```
radu-mocanu, line 21: there is no `datetime` field in `ActionSchema` object

src/uipath/models/attachment.py
```diff
@@ -17,9 +17,14 @@ class Attachment(BaseModel):
         use_enum_values=True,
         arbitrary_types_allowed=True,
         extra="allow",
-        json_encoders={datetime: lambda v: v.isoformat() if v else None},
     )
 
+    @field_serializer("*", when_used="json")
```
radu-mocanu, line 22:
```suggestioncreation_time
    @field_serializer("creation_time", when_used="json")
```
for models that have just one datetime field maybe we can target that specific value, instead of applying the serializer on all fields

ionmincu (approved):

PR #574, opened 2025-09-13
refactor(RuntimeUpdates): decouple runtime from being constructed from a class method

akshaylive:
This will allow for programmatic execution of arbitrary functions without having to rely on scripts.

Additionally, decoupling `entrypoint` from context as it's only applicable for executing scripted agents.


akshaylive:
Closing this as it violates backwards compatibility. We can revisit this in the next major revision.

PR #588, opened 2025-09-16
refactor(ProjectStructure): project structure inheritance

akshaylive:
Project Structure *is a* Project Folder.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.1.53.dev1005881135",

  # Any version from PR
  "uipath>=2.1.53.dev1005880000,<2.1.53.dev1005890000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```


src/uipath/_cli/_utils/_studio_project.py
```diff
@@ -174,6 +148,33 @@ def get_folder_by_name(
     return None
 
 
+def resolve_path(
```
radu-mocanu, line 151: is this method used anywhere?

akshaylive: Nope. This is just a useful method and after speaking with Pufu, it's better to keep it in the base library.

PR #602, opened 2025-09-19
feat(AgentDefinitions): adding utils to load usable agent definitions from cloud

akshaylive:
Specifically, allow loading of low-code agents along with usable tool definitions.


cristipufu:
We have these models here, which we should remove/move/converge https://github.com/UiPath/uipath-langchain-python/blob/main/src/uipath_langchain/builder/agent_config.py

Let's also add these tests https://github.com/UiPath/uipath-langchain-python/blob/main/tests/builder/test_config.py

PR #619, opened 2025-09-25
Add initial version of revamped coded evaluators

andrei-rusu:


src/uipath/eval/coded_evaluators/contains_evaluator.py
```diff
@@ -0,0 +1,55 @@
...
+    """Evaluator that checks if the actual output contains the expected output.
+
+    This evaluator returns True if the actual output contains the expected output,
+    and False otherwise. It supports case sensitivity and negation options.
+    """
+
+    async def evaluate(
+        self,
+        agent_execution: AgentExecution,
+        evaluation_criteria: OutputEvaluationCriteria,
```
Chibionos, line 29: Evaluation Criteria would need to be defined for each Evaluator

Contains will have searchText we don't have a type representing that right now.

Chibionos (changes requested):
Need to standardize the types for the critterias

src/uipath/eval/coded_evaluators/llm_as_judge_evaluator.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+        return self
+
+    def model_post_init(self, __context: Any) -> None:
+        """Initialize the LLM service if not provided."""
+        super().model_post_init(__context)
+        if self.llm_service is None:
+            self.llm_service = self._get_llm_service()
+
+    def _get_llm_service(self):
```
Chibionos, line 68: do we need to have the import from UiPath inside the evaluator ?

andrei-rusu: No, that is there just so it can work with the SDK in it's current form, the method can then be edited with a different default when we move it out. It's isolated now, that's what's important, so it won't be a runtime issue even if you move it as-is, but obviously the cleanest would be to just give it a different default once it's out.

radu-mocanu:
@andrei-rusu let s fix the tests, squash everything into 1 commit and merge this

src/uipath/_cli/cli_run.py
```diff
@@ -119,6 +120,8 @@ def run(
 
         except UiPathRuntimeError as e:
             console.error(f"{e.error_info.title} - {e.error_info.detail}")
+        except UiPathEvaluationError as e:
```
radu-mocanu, line 123: no need for this here

PR #642, opened 2025-10-03
feat(Mocking): allow functions to be marked as mockable

akshaylive:
Initial design wanted to let every tracable elements to also be mockable. However, langgraph and `@traced` annotations uses different tracing mechanisms. Keeping `@mockable` separate from the tracing system will create a cleaner separation of concern.

The `mockingStrategy` can be extended to support different types of strategies. Currently, LLM strategy and mockito strategy is supported.

The LLM strategy is backwards compatible, however the feature is incomplete:

1. The prompt used to generate LLM responses need to be 1:1 with the c# world. Specially, the current run history may be an important part of the prompt.
2. Dynamic input generation is currently unsupported.
3. Model selection is currently unsupported.
4. Caching is currently unsupported.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.1.73.dev1006421502",

  # Any version from PR
  "uipath>=2.1.73.dev1006420000,<2.1.73.dev1006430000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```


src/uipath/eval/mocks/llm_mocker.py
```diff
@@ -0,0 +1,153 @@
...
+                        "arguments": params.get(
+                            "arguments"
+                        ),  # arguments could be passed into tool
+                        "settings": params.get(
+                            "settings"
+                        ),  # settings could be passed into tool
+                        "inputSchema": params.get("input_schema"),
+                    },
+                    "toolSchema": params.get("input_schema"),
+                    "currentToolInput": {
```
bai-uipath, line 117: Do we validate the input anywhere? We probably should return a response to the agent making it clear that the inputs don't match the input schema if this happens, rather than a generic exception.

src/uipath/eval/mocks/llm_mocker.py
```diff
@@ -0,0 +1,153 @@
...
+                        "args": args,
+                        "kwargs": kwargs,
+                    },
+                    "agentInfo": {  # This is incomplete
+                        # "agentName": self.evaluation_item.name, # to be obtained.
```
bai-uipath, line 122: will the agent prompt be passed in here?

src/uipath/eval/mocks/llm_mocker.py
```diff
@@ -0,0 +1,153 @@
...
+Your task is to provide a realistic response for the tool based on its schema, examples, and context.
+
+SELECTED TOOL INFORMATION:
+{toolInfo}
+
+SELECTED TOOL SCHEMA:
+{toolSchema}
+
+SELECTED TOOL EXAMPLES:
+{toolRunExamples}
```
bai-uipath, line 22: I'm really curious how well this works at mocking tool outputs reliably in this design, where we're just generating based on schemas and simulation but without few-shot example grounding. Questions worth validating:

- Semantics of the generated output - will it generate formally correct but nonsensical output, or realistic outputs?

- How consistent is this across runs with the same or similar inputs - will this add variance to evaluations?

- not sure if we'll concatenate the agent prompt as context to this LLM call, but will it get confused by a long agent prompt?

src/uipath/eval/mocks/llm_mocker.py
akshaylive: Yes. [Here's](https://github.com/UiPath/Agents/blob/b0bf7d94a044683b1f9ed2b7560159f17d25bf35/backend/Execution.Shared/Tools/Mocking/LLMInputMocking.cs#L18) the C# code for reference

akshaylive: ^^ My bad, that's for the input generation. For the tool response generation, the agent definition is [this](https://github.com/UiPath/Agents/blob/b0bf7d94a044683b1f9ed2b7560159f17d25bf35/backend/Execution.Shared/Tools/Mocking/LLMMockedToolExecutor.cs#L25). The definition does [not contain](https://github.com/UiPath/Agents/blob/b0bf7d94a044683b1f9ed2b7560159f17d25bf35/backend/Common/Traces/TracesMetadata.cs#L3) the agent prompt.

src/uipath/eval/mocks/llm_mocker.py
akshaylive: Currently there is no few shot grounding. We will be adding it later for 1:1 parity with the C# world. Looks like the `toolRunExamples` are [fetched from tool definition](https://github.com/UiPath/Agents/blob/b0bf7d94a044683b1f9ed2b7560159f17d25bf35/backend/Execution.Shared/Tools/Mocking/SimulatedTool.cs#L47) (under `exampleCalls`).

* While we do not have quantitative performance of this, it seems to work quite well, which is why it was pushed to production. It generates reasonable outputs.
* No idea. I don't think this is in scope for this PR (or related PRs). We are first migrating to coded agents for feature-parity. Once we have parity, we can switch impl of lowCode agents to coded agents. I do think this is a good question to ask, but I'm not sure about the priority of this compared our other tasks.
* As mentioned in the previous comment, agent prompt is not passed.

src/uipath/_cli/_evals/_models/_evaluation_set.py
```diff
@@ -9,28 +9,127 @@ class EvaluationSimulationTool(BaseModel):
     name: str = Field(..., alias="name")
 
 
+class MockingStrategyType(str, Enum):
+    LLM = "llm"
+    MOCKITO = "mockito"
```
radu-mocanu, line 14: https://github.com/UiPath/uipath-python/pull/642/files#r2409715214

src/uipath/_cli/_evals/mocks/mocks.py
```diff
@@ -0,0 +1,140 @@
...
+                )
+                return await func(*args, **kwargs)
+    else:
+
+        @functools.wraps(func)
+        def decorated_func(*args, **kwargs):
+            try:
+                return run_coroutine(mock_response_generator(*args, **kwargs))
+            except Exception:
+                logger.warning(
```
radu-mocanu, line 96: same here

src/uipath/_cli/_evals/mocks/mocks.py
```diff
@@ -0,0 +1,140 @@
...
+        return await mocker.response(func, params, *args, **kwargs)
+
+
+_event_loop = None
+logger = logging.getLogger(__name__)
+
+
+def run_coroutine(coro):
+    """Run a coroutine synchronously."""
+    global _event_loop
```
radu-mocanu, line 54: why do we need a persistent loop here? if the loop dies won't the subsequent calls fail silently?

src/uipath/_cli/_evals/mocks/mocks.py
```diff
@@ -0,0 +1,140 @@
...
+    if not _event_loop:
```
radu-mocanu, line 55:
```suggestion
    if not _event_loop or not _event_loop.is_running():
```

src/uipath/_cli/_evals/mocks/llm_mocker.py
```diff
@@ -0,0 +1,153 @@
...
+Here are the instructions for this run:
+{testRunProctorInstructions}
+
+Based on the above information, provide a realistic response for this tool call. Your response should:
+1. Match the expected output format according to the tool schema
+2. Be very consistent with how the tool has responded in previous examples. Do no omit fields or properties.
+3. Always include the entire output regardless of token length.
+3. Consider the context of the current test run and the agent being tested.  If the agent is acting on a property, make sure the output includes that property.
+
+Respond ONLY with valid JSON that would be a realistic and completetool response. Do not include any explanations or markdown.
```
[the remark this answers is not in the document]
akshaylive: Keeping this in sync with the C# implementation.

src/uipath/_cli/_evals/mocks/mocks.py
```diff
@@ -0,0 +1,140 @@
...
+def mocked_response_decorator(func, params: dict[str, Any]):
+    """Mocked response decorator."""
+
+    async def mock_response_generator(*args, **kwargs):
+        mocked_response = await get_mocked_response(func, params, *args, **kwargs)
+        return_type: Any = func.__annotations__.get("return", None)
+
+        if return_type is not None:
+
+            class Output(BaseModel):
```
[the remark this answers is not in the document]
akshaylive: Good catch!

src/uipath/_cli/_evals/mocks/mocks.py
```diff
@@ -0,0 +1,140 @@
...
+
+async def get_mocked_response(
+    func: Callable[[Any], Any], params: dict[str, Any], *args, **kwargs
+) -> Any:
+    """Get a mocked response."""
+    mocker = mocker_context.get()
+    evaluation_item = evaluation_context.get()
+    if mocker is None or evaluation_item is None:
+        # TODO raise a new UiPath exception type
+        raise RuntimeError(f"Evaluation item {func.__name__} has not been evaluated")
```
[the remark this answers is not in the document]
akshaylive: Sure thing. I can take this in a subsequent PR, after the other is in.

src/uipath/_cli/_evals/mocks/mocks.py
```diff
@@ -0,0 +1,140 @@
...
+
+    is_async = inspect.iscoroutinefunction(func)
+    if is_async:
+
+        @functools.wraps(func)
+        async def decorated_func(*args, **kwargs):
+            try:
+                return await mock_response_generator(*args, **kwargs)
+            except Exception:
+                logger.warning(
```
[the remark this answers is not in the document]
akshaylive: Yea, I'll need to rewrite the exception handling part. Right now, if a mock is not set in context, the `get_mocked_response` throws an exception. This will mean that `@mockable()` code is fail when it's not in evaluation context. This will get easier once the `UiPathEvaluationError` is merged in.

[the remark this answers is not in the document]
akshaylive: `UiPathEvaluationError` is not the right error class. Maybe handle something like `UiPathNoMockError`

src/uipath/_cli/_evals/mocks/mocks.py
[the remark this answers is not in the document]
akshaylive: As spoken over call, this will raise `UiPathNoMockError` or something similar

PR #647, opened 2025-10-06
feat: progress on parallelization of eval runs

mathurk:
- This is the work done so far on parallelization
- Demo: https://uipath-product.slack.com/archives/C093V8BMJTC/p1759425295314939

Did a test with 10 evaluations following config:

image

and varied the number of workers and found this as the general time trends:
image

the time gained is pretty dismal, but something is better than nothing for starters ?

### DIFF
--- a/src/uipath/_cli/_evals/_runtime.py
+++ b/src/uipath/_cli/_evals/_runtime.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import logging
 import uuid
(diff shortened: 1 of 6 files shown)


src/uipath/_cli/_evals/_runtime.py
```diff
@@ -390,3 +462,317 @@ async def cleanup(self) -> None:
...
+
+        self.progress_tracker = ProgressTracker(
+            execution_id=self.context.execution_id or "",
+            total_evals=len(evaluation_set.evaluations),
+            total_batches=len(batches),
+            event_bus=self.event_bus,
+        )
+
+        tasks = []
+        for batch in batches:
```
akshaylive, line 570: This is pretty inefficient as a single evaluation can significantly slow down the batch processing.

My suggestion is to completely ditch the batch concept. I.e; just submit individual evaluation item to `execute_evaluation_with_semaphore`. In other words, always use batch size=1 and num_workers = batch_size.

Issue #635, opened 2025-10-01
Invoking `chat_completions` with `response_format=pydantic_model` throws exception


akshaylive:
Here's a simple reproduction step:

```
llm = UiPath().llm
class OutputFormat(BaseModel):
  result: str
response = await llm.chat_completions([("human", f"repeat this: hi!")], response_format=OutputFormat)
```

Relevant exception:
```
    response = await llm.chat_completions([("human", f"repeat this: hi!")], response_format=OutputFormat)
  File "uipath-python/src/uipath/tracing/_traced.py", line 470, in decorated_func
    return await func(*args, **kwargs)
  File "uipath-python/src/uipath/tracing/_traced.py", line 263, in async_wrapper
    inputs = _SpanUtils.format_args_for_trace_json(
  File "uipath-python/src/uipath/tracing/_utils.py", line 277, in format_args_for_trace_json
    return json.dumps(result, default=_simple_serialize_defaults)
...
  File "uipath-python/src/uipath/tracing/_utils.py", line 22, in _simple_serialize_defaults

```

PR #648, opened 2025-10-07
fix(response_format): added role mapping + changed default serialization

giuliastf:
**Description**
* Updated the `chat_completions` method in `llm_gateway_service.py` to accept both lists of dictionaries and lists of tuples as message input, normalizing roles like "human" and "ai" to "user" and "assistant" for consistency.

* Improved the `_simple_serialize_defaults` function in `tracing/_utils.py` to handle Pydantic `BaseModel` instances and classes, including schema serialization for classes and preventing serialization of class types as instances.
**Issue** https://github.com/UiPath/uipath-python/issues/635


src/uipath/tracing/_utils.py
```diff
@@ -1,321 +1,331 @@
...
+            return obj._asdict()
+        return list(obj)
+
+    if isinstance(obj, datetime):
+        return obj.isoformat()
+
+    if isinstance(obj, (timezone, ZoneInfo)):
+        return obj.tzname(None)
+
+    return str(obj)
```
akshaylive, line 57: Shouldn't we allow primitives `None`, `int`, `float`, `bool` to pass through without it being considered a string?

PR #653, opened 2025-10-07
feat(Documents): add support for extract from modern projects

alexandrujircan:
### Description
Modify `extract` operation to also support DU Modern projects.

### Docs:
- Added more examples

### Tests:
- Coverage 100%

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.1.111"
+version = "2.1.112"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.10"


src/uipath/_services/documents_service.py
```diff
@@ -37,11 +38,13 @@ def _get_common_headers(self) -> Dict[str, str]:
...
 
-    def _get_project_id_by_name(self, project_name: str) -> str:
+    def _get_project_id_by_name(
+        self, project_name: str, project_type: ProjectType
+    ) -> str:
         response = self.request(
             "GET",
             url=Endpoint("/du_/api/framework/projects"),
-            params={"api-version": 1.1, "type": "IXP"},
+            params={"api-version": 1.1},
```
vlupoaie, line 47: `project_type` needs to be used here, right?

src/uipath/_services/documents_service.py
```diff
@@ -143,16 +154,79 @@ async def _start_digitization_async(
...
+    def _get_document_type_id(
+        self,
+        project_id: str,
+        document_type_name: Optional[str],
+        project_type: ProjectType,
+    ) -> str:
+        if project_type == ProjectType.IXP:
+            return str(UUID(int=0))
+        if project_type == ProjectType.MODERN and document_type_name is None:
+            raise ValueError(
```
alexandrujircan: Maybe I’ll remove this since it’s a private method, to make sure it’s called correctly.

vlupoaie: Yeah, this should be handled higher up, right?

alexandrujircan: Yes

PR #654, opened 2025-10-07
docs: add instruction to activate the venv

mjnovice:


akshaylive (changes requested):
Thanks for fixing the documentation! ❤️

PR #695, opened 2025-10-14
feat(InputMocking): generate mock input using LLM given instructions

bai-uipath:
This PR adds LLM-based input mocking, using prompt configured in `inputMockingStrategy` in the evaluation JSON.

Currently missing the input schema (which we will extract from the agent definition, once it is available).

### DIFF
--- a/src/uipath/_cli/_evals/_runtime.py
+++ b/src/uipath/_cli/_evals/_runtime.py
@@ -11,6 +11,10 @@ from opentelemetry import context as context_api
 from opentelemetry.sdk.trace import ReadableSpan, Span
 from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
 
+from uipath._cli._evals.mocks.input_mocker import (
+    generate_llm_input,
+)
+
 from ..._events._event_bus import EventBus
 from ..._events._events import (
     EvalItemExceptionDetails,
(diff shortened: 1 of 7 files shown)


src/uipath/_cli/_evals/_runtime.py
```diff
@@ -417,6 +427,18 @@ async def _execute_eval(
...
         return evaluation_run_results
 
+    async def _generate_input_for_eval(
+        self, eval_item: EvaluationItem
+    ) -> EvaluationItem:
+        """Use LLM to generate a mock input for an evaluation item."""
+        if not self.context.entrypoint:
+            raise UiPathInputMockingError("entrypoint must be provided for eval runs")
+
+        input_schema = extract_input_schema_from_entrypoint(self.context.entrypoint)
```
akshaylive: Not sure this is a good idea. It's better to extract from the agent definition. Can we somehow ensure the evaluation starts with an `AgentDefinition`? Maybe we want a `CodedAgentDefinition` (similar to `LowCodeAgentDefinition`)

src/uipath/_cli/_evals/mocks/input_mocker.py
```diff
@@ -0,0 +1,149 @@
...
+            adapter = TypeAdapter(first_param.annotation)
+            return adapter.json_schema()
+
+    raise UiPathInputMockingError(
+        f"No suitable entrypoint (main, run, execute) with typed parameters found in {entrypoint_path}"
+    )
+
+
+@traced(name="__mocker__")
+async def generate_llm_input(
```
akshaylive: Let's add some tests for this?

akshaylive (approved):
This line need not exist right?
```
@pytest.mark.httpx_mock(assert_all_responses_were_requested=False)
```

Otherwise, LGTM!

PR #703, opened 2025-10-15
feat: add support for custom evaluators

radu-mocanu:
- add support for custom evaluators
- refactored coded evaluator files pushing to deduplicate code
- push custom evaluators (.py files) and evaluators types files

# Add and register custom evaluator

1. (Optional) Add a new evaluator -> can be created manually in the evals/custom-evaluators directory
```
uipath add evaluator my_custom_evaluator
```
2. Implement the logic

3. Register the evaluator
```
uipath register evaluator my_custom_evaluator
```
4. Apply it to any dataset

> The custom evaluator is self registering when the code is pushed

### DIFF
--- a/samples/calculator/evals/evaluators/correct-operator-evaluator.json
+++ b/samples/calculator/evals/evaluators/correct-operator-evaluator.json
@@ -1,0 +1,14 @@
+{
+  "version": "1.0",
+  "id": "CorrectOperatorEvaluator",
+  "evaluatorTypeId": "file://types/correct-operator-evaluator-types.json",
+  "evaluatorSchema": "file://correct_operator.py:CorrectOperatorEvaluator",
+  "description": "A custom evaluator that checks if the correct operator is being used by the agent ",
+  "evaluatorConfig": {
+    "name": "CorrectOperatorEvaluator",
+    "defaultEvaluationCriteria": {
+      "operator": "+"
+    },
+    "negated": false
+  }
+}
--- a/samples/calculator/main.py
+++ b/samples/calculator/main.py
@@ -39,6 +39,9 @@ async def get_random_operator() -> Wrapper:
     """Get a random operator."""
     return Wrapper(result=random.choice([Operator.ADD, Operator.SUBTRACT, Operator.MULTIPLY, Operator.DIVIDE]))
 
+@traced(name="track_operator")
+def track_operator(operator: Operator):
+    pass
 
 @traced()
 async def main(input: CalculatorInput) -> CalculatorOutput:
@@ -46,6 +49,7 @@ async def main(input: CalculatorInput) -> CalculatorOutput:
         operator = (await get_random_operator()).result
     else:
         operator = input.operator
+    track_operator(operator)
     match operator:
         case Operator.ADD: result = input.a + input.b
         case Operator.SUBTRACT: result = input.a - input.b
--- a/src/uipath/_cli/_evals/_helpers.py
+++ b/src/uipath/_cli/_evals/_helpers.py
@@ -1,0 +1,194 @@
+# type: ignore
+import ast
+import importlib.util
+import json
+import logging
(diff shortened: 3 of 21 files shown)


samples/calculator/main.py
```diff
@@ -39,13 +39,17 @@ async def get_random_operator() -> Wrapper:
     """Get a random operator."""
     return Wrapper(result=random.choice([Operator.ADD, Operator.SUBTRACT, Operator.MULTIPLY, Operator.DIVIDE]))
 
+@traced(name="track_operator")
```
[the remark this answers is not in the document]
radu-mocanu: adds a span to the agent trace, we use it as part of the custom evaluator logic https://github.com/UiPath/uipath-python/pull/703/files#diff-44ea7762c07ffd2865a9305251a9ddf8dd689e9aa46452cb4ae8647651d0a9a6R24

src/uipath/_cli/_evals/_helpers.py
```diff
@@ -0,0 +1,194 @@
...
+        return None
+
+    file_path = custom_evaluators_path / filename
+    if file_path.exists():
+        return file_path
+
+    return None
+
+
+def find_base_evaluator_class(file_path: Path) -> Optional[str]:
```
akshaylive: This basically fails if there are more than one evaluator. Should we support something like `foo.py:Bar` where `Bar` is the classname?

src/uipath/_cli/_evals/_helpers.py
```diff
@@ -0,0 +1,194 @@
...
+    finally:
+        # Remove from sys.path
+        if parent_dir in sys.path:
+            sys.path.remove(parent_dir)
+
+
+def generate_evaluator_config(evaluator_class: type, class_name: str) -> dict[str, Any]:
+    """Generate the evaluator config from the class."""
+    try:
+        config_type = evaluator_class._extract_config_type()
```
akshaylive: Could you explain this logic please? `_extract_config_type` is not valid for all types right? Should we update the type def for `evaluator_class`?

samples/calculator/evals/evaluators/correct-operator-evaluator.json
```diff
@@ -0,0 +1,14 @@
+{
+  "version": "1.0",
+  "id": "CorrectOperatorEvaluator",
+  "evaluatorTypeId": "file://types/correct-operator-evaluator-types.json",
+  "evaluatorSchema": "file://correct_operator.py:CorrectOperatorEvaluator",
```
akshaylive: `file://` typically starts with root. Regardless, should we point this to an import module like `module.submodule:ClassName`?

src/uipath/_cli/_evals/_helpers.py
```diff
@@ -0,0 +1,194 @@
...
+
+def register_evaluator(filename: str) -> tuple[str, str]:
+    """Infers the schema and types of a custom evaluator
+
+    Returns:
+        tuple[str, str]:
+            - The first string is the path to the python evaluator file.
+            - The second string is the evaluator type that corresponds to the schema file.
+    """
+    if not filename.endswith(".py"):
```
mjnovice: should we not fail with an error if the file is not a python file (since this is a python SDK and we don't support other languages at the moment) ?

src/uipath/_cli/_evals/_helpers.py
radu-mocanu: @andrei-rusu worked on this part. the same method is used to generate types for platform evaluators too 
https://github.com/UiPath/uipath-python/blob/release/revamped-evals/src/uipath/eval/evaluators_types/generate_types.py#L20

akshaylive: What I mean is, should we change 
```
evaluator_class: type[BaseEvaluator]
```

PR #706, opened 2025-10-15
refactor(CLIPull): moving project download to utils

akshaylive:
This is better for code reuse.


cotovanu-cristian (approved):
[nit] not in the scope of this PR necessarily but we should think if `_project_files.py` is worth refactoring a bit as it's getting bigger and bigger 🤔

PR #710, opened 2025-10-16
wip feat(Documents): add support for digitization

alexandrujircan:


alexandrujircan:
We have decided that, for now, we will not expose the digitize operation in the UiPath Python SDK.

PR #724, opened 2025-10-17
feat: missing changes from llm eval wiring

mjnovice:


src/uipath/eval/coded_evaluators/llm_judge_output_evaluator.py
```diff
@@ -81,7 +83,7 @@ class LLMJudgeOutputEvaluator(BaseLLMOutputEvaluator[LLMJudgeOutputEvaluatorConf
     @classmethod
     def get_evaluator_id(cls) -> str:
         """Get the evaluator id."""
-        return "uipath-llm-judge-output-semantic-similarity"
+        return EvaluatorType.LLM_JUDGE_OUTPUT_SEMANTIC_SIMILARITY.value
```
andrei-rusu: Shall we do this for all the remaining evaluators?

mjnovice: yes. would make sense.

Chibionos (approved):

PR #748, opened 2025-10-21
feat: add agent loop flow control tools

cotovanu-cristian:
## What changed?
- add uipath react loop flow control tools so they can be reused cross agent framework SDKs
- fix the process tool types as they were capitalized in the agent nupkg and allow both lowercase and capitalized types


src/uipath/agent/models/agent.py
```diff
@@ -52,11 +52,11 @@ class BaseAgentToolResourceConfig(BaseAgentResourceConfig):
 class AgentToolType(str, Enum):
     """Agent tool type."""
 
-    AGENT = "agent"
-    PROCESS = "process"
-    API = "api"
-    PROCESS_ORCHESTRATION = "processorchestration"
-    INTEGRATION = "integration"
+    AGENT = "Agent"
```
akshaylive: IIRC, the lowcode `resource` definitions are lowercased, right?

cotovanu-cristian: in the agent package they are not, that's the problem.
what we can do is to lowercase before deserialization through a field_validator and accept both formats -- similar to what we do for `EscalationRecipientType`

PR #750, opened 2025-10-21
feat: refactor runtime system for extensible multi-entrypoint support

bai-uipath:
This PR refactors the runtime architecture to support multiple entrypoints in a project and a pluggable runtime factory system using `Middlewares.register_runtime_factory()` to retrieve a generic runtime factory and updates init, run, and eval to use the new runtime factory.

The corresponding change in the langchain repo registers the runtime factory for langgraph scripts: https://github.com/UiPath/uipath-langchain-python/pull/231

### DIFF
--- a/src/uipath/_cli/_runtime/_runtime.py
+++ b/src/uipath/_cli/_runtime/_runtime.py
@@ -76,16 +76,16 @@ class UiPathRuntime(UiPathBaseRuntime):
 console = ConsoleLogger()
 
 
-def get_user_script(directory: str, entrypoint: Optional[str] = None) -> Optional[str]:
-    """Find the Python script to process."""
+def get_user_scripts(directory: str, entrypoint: Optional[str] = None) -> List[str]:
+    """Find all Python scripts to process."""
     if entrypoint:
         script_path = os.path.join(directory, entrypoint)
         if not os.path.isfile(script_path):
             console.error(
                 f"The {entrypoint} file does not exist in the current directory."
             )
-            return None
-        return script_path
+            return []
+        return [script_path]
 
     python_files = [f for f in os.listdir(directory) if f.endswith(".py")]
 
@@ -93,14 +93,9 @@ def get_user_script(directory: str, entrypoint: Optional[str] = None) -> Optiona
         console.error(
             "No python files found in the current directory.\nPlease specify the entrypoint: `uipath init <entrypoint_path>`"
         )
-        return None
-    elif len(python_files) == 1:
-        return os.path.join(directory, python_files[0])
-    else:
-        console.error(
-            "Multiple python files found in the current directory.\nPlease specify the entrypoint: `uipath init <entrypoint_path>`"
-        )
-        return None
+        return []
+
+    return [os.path.join(directory, f) for f in python_files]
 
 
 class UiPathScriptRuntime(UiPathRuntime):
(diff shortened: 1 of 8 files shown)


src/uipath/_cli/_runtime/_runtime.py
```diff
@@ -76,31 +76,26 @@ async def execute(self) -> Optional[UiPathRuntimeResult]:
...
-    elif len(python_files) == 1:
-        return os.path.join(directory, python_files[0])
-    else:
-        console.error(
-            "Multiple python files found in the current directory.\nPlease specify the entrypoint: `uipath init <entrypoint_path>`"
-        )
-        return None
+        return []
+
+    return [os.path.join(directory, f) for f in python_files]
```
akshaylive: Beautiful! :)

However, I do think we need to filter this. The ScriptExecutor does this:
```
        for func_name in ["main", "run", "execute"]:
            if hasattr(module, func_name):
```

so let's ensure that the files have these defined, yea?

src/uipath/_cli/_runtime/_runtime_factory.py
```diff
@@ -1,21 +1,53 @@
...
+
+        runtimes = []
+        for script_path in scripts:
+            runtime = self._create_runtime(script_path)
+            if runtime:
+                runtimes.append(runtime)
+
+        return runtimes
+
+    def get_runtime(self, entrypoint: str) -> Optional[UiPathScriptRuntime]:
```
akshaylive: As mentioned above this method is redundant -- can be safely removed

src/uipath/_cli/cli_eval.py
```diff
@@ -138,7 +137,9 @@ def eval(
         asyncio.run(console_reporter.subscribe_to_eval_runtime_events(event_bus))
 
         try:
-            runtime_factory = generate_runtime_factory()
+            # Last one is the Python script runtime factory
+            runtime_factory = Middlewares._runtime_factories[-1]
```
akshaylive: This does not have the fallback logic..

Know what i just realized?
```
class FallbackRuntimeFactory(UiPathBaseRuntimeFactory):
  def __init__(self, ..):
    self. factories = []

  def from_context(...):
    for factory in self.factories:
       try:
          return factory.from_context(..)
       except:
          ....

  def discover_...
    # this does the merging logic..
```

src/uipath/_cli/_runtime/_runtime_factory.py
```diff
@@ -1,21 +1,53 @@
...
+        """Get runtime for a specific Python script."""
+        if not os.path.isabs(entrypoint):
+            script_path = os.path.abspath(entrypoint)
+        else:
+            script_path = entrypoint
+
+        if not os.path.isfile(script_path) or not script_path.endswith(".py"):
+            return None
+
+        return self._create_runtime(script_path)
```
akshaylive: As mentioned above, this should also have one of the three special functions defined.

src/uipath/_cli/cli_eval.py
akshaylive: What I mean is, if someone imports langgraph library but doesn't define a graph,`Middlewares._runtime_factories[-1]` will end up using the langgraph factory and will never fallback to the script runtimefactory.

PR #773, opened 2025-10-24
refactor(RemoveAny): remove any types

akshaylive:
Removing Any* types to make the code maintainable in the long run.


src/uipath/_cli/_evals/_progress_reporter.py
```diff
@@ -132,7 +131,9 @@ def _get_endpoint_prefix(self) -> str:
             return "api/"
         return "agentsruntime_/api/"
 
-    def _is_coded_evaluator(self, evaluators: List[AnyEvaluator]) -> bool:
+    def _is_coded_evaluator(
+        self, evaluators: List[BaseEvaluator[Any, Any, Any]]
```
radu-mocanu: this method will always categorize evaluators as `coded`. might break reporting

Chibionos: I think for coded agents the legacy evaluators should not be supported anymore we should only support the new ones

mjnovice: decided to make it backward compatible

radu-mocanu (approved):

PR #787, opened 2025-10-25
feat(ReturnAgentExecution): return agent execution details as part of evaluation output

akshaylive:
Specifically,
1. adding `agent_execution_output` to `EvaluationRunResult`
2. keeping `evaluatorId` in `EvaluationRunResultDto`, so that the object can be reconstructed after serialization.
3. refactored entrypoint to be a cached property so that it's evaluated lazily.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.1.113.dev1007872222",

  # Any version from PR
  "uipath>=2.1.113.dev1007870000,<2.1.113.dev1007880000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### DIFF
--- a/samples/calculator/evals/eval-sets/legacy.json
+++ b/samples/calculator/evals/eval-sets/legacy.json
@@ -21,7 +21,7 @@
       "expectedOutput": {
         "result": 2
       },
-      "expectedAgentBehavior": "",
+      "expectedAgentBehavior": "The operation should produce the right output.",
       "evalSetId": "default-eval-set-id",
       "createdAt": "2025-09-04T18:54:58.378Z",
       "updatedAt": "2025-09-04T18:55:55.416Z"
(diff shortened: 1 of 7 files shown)


src/uipath/_cli/_evals/_runtime.py
```diff
@@ -174,6 +176,13 @@ def __init__(
         self.logs_exporter: ExecutionLogsExporter = ExecutionLogsExporter()
         self.execution_id = str(uuid.uuid4())
 
+    @cached_property
+    async def get_entrypoint(self):
```
bai-uipath: Interesting that this needs to be loaded lazily - I thought all runtimes have an entrypoint. Curious, when is this not the case?

samples/calculator/evals/eval-sets/legacy.json
```diff
@@ -21,7 +21,7 @@
       "expectedOutput": {
         "result": 2
       },
-      "expectedAgentBehavior": "",
+      "expectedAgentBehavior": "The operation should produce the right output.",
```
bai-uipath: Probably better to have a more specific text here about what the agent should be doing.

PR #800, opened 2025-10-28
fix: update pull to use coded-evals folder and add conflict detection

Chibionos:
## What's Changed

This PR fixes two issues with the `uipath` CLI commands:

### 1. Pull Command Now Uses `coded-evals` Folder

**Problem**: The `uipath pull` command was downloading from the `evals/` folder, but `uipath push` uploads to `coded-evals/`. This meant files pushed weren't being pulled back.

**Solution**: Updated pull to download from `coded-evals/` folder, matching where push uploads files.

### 2. Added Interactive Conflict Detection

**Problem**: Both push and pull would silently overwrite files without warning users about conflicts.

**Solution**: Added interactive prompts that:
- Detect when local and remote files differ
- Show the local file's modification timestamp
- Ask users whether to overwrite or skip
- Use appropriate messaging for push vs pull operations

### Example User Experience

When pulling with conflicts:
```
File 'main.py' has different content on remote.
Local file last modified: 2025-10-28 04:53:14
Do you want to overwrite the local file with the remote version? [y/N]:
```

When pushing with conflicts:
```
File 'main.py' has different content on remote.
Local file last modified: 2025-10-28 04:57:23
Do you want to push and overwrite the remote file with your local version? [y/N]:
```

## Testing

Added comprehensive unit tests:
- ✅ 11 tests for conflict handlers (InteractiveConflictHandler, PushConflictHandler)
- ✅ 4 tests for push conflict detection
- ✅ 2 new tests for pull with coded-evals folder
- ✅ Updated existing pull tests to reflect new behavior

All tests passing.

## Version Bump

Bumped version from `2.1.117` → `2.1.118`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.1.128"
+version = "2.1.129"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.10"
--- a/src/uipath/utils/dynamic_schema.py
+++ b/src/uipath/utils/dynamic_schema.py
@@ -1,17 +1,20 @@
 """Json schema to dynamic pydantic model."""
 
-from typing import Any, Dict, List, Optional, Type, Union
+from enum import Enum
+from typing import Any, Dict, List, Type, Union
 
 from pydantic import BaseModel, Field, create_model
 
 
 def jsonschema_to_pydantic(
     schema: dict[str, Any],
-    definitions: Optional[dict[str, Any]] = None,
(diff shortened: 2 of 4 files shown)


src/uipath/_cli/_utils/_project_files.py
```diff
@@ -76,11 +84,143 @@ class AlwaysSkipHandler:
...
+
+        # Prompt user for confirmation
+        sys.stdout.flush()
+        return click.confirm(
+            "Do you want to overwrite the local file with the remote version?",
+            default=False,
+        )
+
+
+class PushConflictHandler:
```
radu-mocanu: this class is an unnecessary duplication of `InteractiveConflictHandler`.
the logic is the same, just messages differ. instead we can add an instace variable to InteractiveConflictHandler that specifies if we are in a push or pull context

PR #803, opened 2025-10-28
feat(SupportNonRecordingTraces): support non-recording traces

akshaylive:
Customers can disable tracing in spans by using recording flag. Children of NonRecordingSpans are also NonRecording. By taking this approach, LLM judges do not see mocking internal spans.

To use this, just mark the traced attribute with `recording=False`.

```
@traced(name="bar", recording=True)
def bar():
    return

@traced(name="foo", recording=False)
def foo():
    bar()
```

If a method invokes `foo()`, none of the traces (neither `foo` nor `bar`) will be captured.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.1.122.dev1008032340",

  # Any version from PR
  "uipath>=2.1.122.dev1008030000,<2.1.122.dev1008040000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### DIFF
--- a/src/uipath/tracing/_traced.py
+++ b/src/uipath/tracing/_traced.py
@@ -7,7 +7,7 @@ from functools import wraps
 from typing import Any, Callable, List, Optional, Tuple
 
 from opentelemetry import context, trace
-from opentelemetry.trace import set_span_in_context
+from opentelemetry.trace import NonRecordingSpan, set_span_in_context
 
 from ._utils import _SpanUtils
 
@@ -175,6 +175,7 @@ def _opentelemetry_traced(
     span_type: Optional[str] = None,
     input_processor: Optional[Callable[..., Any]] = None,
     output_processor: Optional[Callable[..., Any]] = None,
+    recording: bool = True,
 ):
     """Default tracer implementation using OpenTelemetry."""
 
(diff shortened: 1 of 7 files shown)


src/uipath/tracing/_utils.py
```diff
@@ -334,34 +334,12 @@ def format_args_for_trace(
             )
             return {"args": args, "kwargs": kwargs}
 
-    @staticmethod
-    def _has_ancestor_with_name(
```
bai-uipath: hmm, do we not need to check the ancestors anymore for the recording tag because they're automatically propagated to children?

[The spec](https://opentelemetry.io/docs/specs/otel/trace/api/#isrecording) says otherwise

> Note that any child span’s recording is determined independently from the value of this flag

may be good to add a unit test if we're relying on this undocumented behavior.

src/uipath/_services/llm_gateway_service.py
```diff
@@ -123,6 +123,10 @@ def clean_type(type_def):
                 properties[key] = clean_type(value)
             cleaned_type["properties"] = properties
 
+        # TODO: what's the best logic to represent this?
```
bai-uipath: this looks like an unrelated comment?

src/uipath/tracing/_traced.py
```diff
@@ -437,6 +450,7 @@ def traced(
             Should accept the function output and return a processed value
         hide_input: If True, don't log any input data
         hide_output: If True, don't log any output data
+        recording: If True, current span all child spans are not captured regardless of their recording status.
```
bai-uipath: I think you meant like

> If False, current span and all child spans are not captured regardless of their recording status

bai-uipath (approved):
lgtm

PR #829, opened 2025-10-31
feat: add assertion for the tools-evals integration tests

mjnovice:


ionmincu (approved):
Alpha seems to fail because of no licenses, will fix later

PR #831, opened 2025-11-01
feat: add live integration tests framework with buckets service

uipreliga:
Implement comprehensive live integration testing framework for the UiPath Python SDK, starting with BucketsService as proof of concept.

### Framework Features
- Function-scoped async fixtures for proper event loop lifecycle management
- Separate sync/async cleanup trackers with automatic resource teardown
- CI-aware credential handling with environment variable fallback
- Pytest markers for granular test selection (live, live_buckets, smoke, slow)
- Default exclusion of live tests to prevent accidental execution

### Test Coverage (21 tests across 4 classes)
- **BasicOperations**: CRUD, list, exists, filtering
- **FileOperations**: upload/download (memory & disk), list with prefix/glob, recursive operations
- **LargeScale**: Iterator efficiency with 20+ files (marked slow)
- **Async**: Async variants of all major operations


.github/workflows/live_tests.yml
```diff
@@ -0,0 +1,101 @@
+name: Live Integration Tests
+
+on:
+  workflow_dispatch: # Manual trigger
```
uipreliga: Some key facts about this approach:
1. The workflow is configured as **manual-only** (`workflow_dispatch`)
2. Scheduled/periodic runs are **commented out** and explicitly disabled
3. The workflow requires explicit user action to trigger
4. Scheduling is an **optional configuration**, not a framework limitation

testcases/apicalls-testcase/main.py
```diff
@@ -1,33 +1,131 @@
...
+                # Delete all files in bucket first
+                files = list(sdk.buckets.list_files(name=resource_id))
+                for file in files:
+                    try:
+                        sdk.buckets.delete_file(name=resource_id, file_path=file.path)
+                    except Exception as file_err:
+                        logger.warning(
+                            f"Could not delete file '{file.path}': {file_err}"
+                        )
+                # Now delete the bucket
```
cosminvlad: No need to do this, when you delete a bucket, all files are deleted too.

PR #848, opened 2025-11-05
fix: agent resources discriminator

cristipufu:
## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.1.140.dev1008482557",

  # Any version from PR
  "uipath>=2.1.140.dev1008480000,<2.1.140.dev1008490000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### DIFF
--- a/src/uipath/agent/models/agent.py
+++ b/src/uipath/agent/models/agent.py
@@ -1,9 +1,11 @@
 """Agent Models."""
 
+from __future__ import annotations
+
 from enum import Enum
 from typing import Annotated, Any, Dict, List, Literal, Optional, Union
 
-from pydantic import BaseModel, ConfigDict, Discriminator, Field, Tag, field_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 from uipath.models import Connection
 from uipath.models.guardrails import (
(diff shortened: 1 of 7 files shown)


src/uipath/agent/models/agent.py
```diff
@@ -645,83 +459,191 @@ class AgentBuiltInValidatorGuardrail(BuiltInValidatorGuardrail):
...
-    resources: List[AgentResourceConfig] = Field(
-        ..., description="List of tools, context, mcp and escalation resources"
-    )
-    features: List[Any] = Field(default_factory=list, description="Agent feature list")
-    settings: AgentSettings = Field(..., description="Agent settings")
 
-    model_config = ConfigDict(
-        validate_by_name=True, validate_by_alias=True, extra="allow"
-    )
+class AgentDefinition(BaseModel):
```
akshaylive: I was thinking we need to remove `messages` , `settings`  etc from there since those are applicable only in low-code scenarios. The base agent definition (could contain only the entrypoint, resources etc) and let lowcode/coded agents etc extend from that.

Also note that Resource is a superset of bindings and contains useful details such as I/O schema. We could deterministically convert from resource to bindings object.

Basically, IMO the AgentDefinition should have a 1:1 mapping with a Runtime and we could send the agent definition to the runtime object during runtime object construction. i.e; `agent = LowCodeAgentRuntime(agent_definition=low_code_agent_definition)`

As far as `LowCode` is concerned, maybe `ReAct` is a better name for this.

cotovanu-cristian (approved):

PR #852, opened 2025-11-06
feat: add is_conversational to uipath.json

GabrielVasilescu04:
# Description

Add support for isConversational. Property should be copied from uipath.json's settings dict, into the operate.json file upon using uipath pack

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.1.140.dev1008522576",

  # Any version from PR
  "uipath>=2.1.140.dev1008520000,<2.1.140.dev1008530000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### DIFF
--- a/src/uipath/_cli/cli_init.py
+++ b/src/uipath/_cli/cli_init.py
@@ -228,6 +228,7 @@ def init(entrypoint: str, infer_bindings: bool, no_agents_md_override: bool) -> 
 
                 config_data = RuntimeSchema(
                     entryPoints=[await runtime.get_entrypoint()],
+                    settings={"isConversational": False},
                 )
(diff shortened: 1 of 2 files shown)


src/uipath/_cli/cli_init.py
```diff
@@ -228,6 +228,7 @@ async def initialize() -> None:
 
                 config_data = RuntimeSchema(
                     entryPoints=[await runtime.get_entrypoint()],
+                    settings={"isConversational": False},
```
radu-mocanu: we need to persist the current configuration when running subsequent inits.

PR #868, opened 2025-11-11
feat(documents): add support for predefined projects

alexandrujircan:
### Description
Add support for predefined projects

### Tests
- coverage 100% + tested locally


vlupoaie (approved):
LGTM. But we should really consider at least changing the `documents_service` module to a package, split the functionality in different modules and then either through composition or inheritance stick them back together in a single `DocumentService` class (as we previously discussed we don't want this one split for now at least).

PR #871, opened 2025-11-11
feat: allow function to be treated as tool via `traced` and modify calculator sample to reflect

andrei-rusu:


cristipufu:
What happens when we need to trace something else from this list?
https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md

We should be **configuring trace semantics**:
```python
@traced(span_kind="tool")
def my_function():
    pass
```

```python
@traced(span_type="TOOL")
def my_function():
    pass
```

PR #877, opened 2025-11-12
chore: remove uv.lock in calculator sample

mjnovice:


Chibionos (approved):

mjnovice:
> don't remove, update it with 'uv sync'

@cristipufu this was added unintentionally in #876 hence removing it.

PR #908, opened 2025-11-18
fix: update sample dependencies to uipath>=2.1.169

Chibionos:
Updates all sample projects to require uipath>=2.1.169 instead of uipath>=2.1.164 due to a breaking change introduced in version 2.1.169.

This ensures all samples use the minimum version that includes the necessary fixes and prevents runtime issues with older versions.

## Updated samples:
- asset-modifier-agent
- calculator
- event-trigger
- google-ADK-agent
- llm_chat_agent
- resource-overrides
- weather_tools


radu-mocanu (commented):
uv lock should also be updated

I'll handle the changes

Issue #669, opened 2025-10-10
Reference doesn't exist on QueueItem Model


Samuel-Ember:
I am looking at creating QueueItems programmatically in python. 
I noticed that you cannot add a reference to allow queue items to be uniquely identified without ignoring types on the QueueItem object. 

I would be nice if this was added as part of the QueueItem model for typing. 

Here is an example of what i have
```py
  queue_item = QueueItem(
  	Name=queue_name,
    # I have to add ignore here for my typing to work
  	Reference="Test Reference", # type: ignore 
  	SpecificContent={
  		"TestKey": "TestValue"
  	}
  )
```

PR #980, opened 2025-12-04
QueueItems: Add Reference property

StefanPopaUi:
# Description
https://github.com/UiPath/uipath-python/issues/669


radu-mocanu (changes requested):
this change requires patch version increase in `pyproject.toml`

PR #1013, opened 2025-12-12
feat: add support for ixp extraction non-blocking

radu-mocanu:
- add support for ixp extraction non-blocking


src/uipath/platform/documents/_documents_service.py
```diff
@@ -888,6 +958,236 @@ async def classify_async(
...
+            )
+        else:
+            file_name = getattr(file, "name", "document")
+            mime_type = mimetypes.guess_type(file_name)[0] or "application/octet-stream"
+            attachment_id = self._attachments_service.upload(
+                name=file_name,
+                content=file,
+            )
+
+        document_id = self._start_digitization_from_attachment(
```
alexandrujircan: Why do we want to use digitization from the attachment? Why not use the document directly?

PR #1028, opened 2025-12-17
fix: legacy evaluation reporting to backend

Chibionos:
## Summary
- Fixed status enum values: Changed from string status ("InProgress") to integer values as backend expects integers
- Added required `source: 0` field to evalSetRun payload for all evaluation types
- Fixed expectedOutput extraction logic to use `is_coded` parameter instead of `isinstance` check
- Fixed evaluator type checking order to check LegacyBaseEvaluator before BaseEvaluator

## Background
Legacy evaluations were failing with 400 errors when reporting results to the backend API. The issues were:

1. Backend's `JsonStringEnumConverter` is disabled, so it expects integer enum values, not strings
2. The `source` field is required by the backend but was only added for coded evaluations
3. Legacy evals are migrated to `EvaluationItem` format with `expectedOutput` inside `evaluationCriterias`, so the `isinstance(eval_item, EvaluationItem)` check always returned True, causing `evaluationCriterias` to be sent instead of `expectedOutput`
4. `LegacyBaseEvaluator` inherits from `BaseEvaluator`, so the isinstance check order matters

## Test plan
- [x] Tested legacy evaluations with calculator sample - all API calls return 200 OK
- [x] Tested coded evaluations with calculator sample - all API calls return 200 OK
- [x] Verified evaluation results are saved to backend

### DIFF
--- a/src/uipath/_cli/_evals/_progress_reporter.py
+++ b/src/uipath/_cli/_evals/_progress_reporter.py
@@ -296,7 +296,6 @@ class StudioWebProgressReporter:
         evaluator_scores: list[dict[str, Any]] = []
 
         for k, v in evaluators.items():
-            # Check LegacyBaseEvaluator first since it inherits from BaseEvaluator
             if isinstance(v, LegacyBaseEvaluator):
                 legacy_evaluators[k] = v
             elif isinstance(v, BaseEvaluator):


src/uipath/_cli/_evals/_progress_reporter.py
```diff
@@ -721,7 +729,7 @@ def _update_eval_run_spec(
         payload: dict[str, Any] = {
             "evalRunId": eval_run_id,
             # For coded evaluations, use integer status; for legacy, use string
-            "status": status.value if is_coded else self._status_to_string(status),
+            "status": status.value,
```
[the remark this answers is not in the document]
mjnovice: +1

src/uipath/_cli/_evals/_progress_reporter.py
```diff
@@ -910,7 +924,7 @@ def _update_eval_set_run_spec(
         payload: dict[str, Any] = {
             "evalSetRunId": eval_set_run_id,
             # For coded evaluations, use integer status; for legacy, use string
-            "status": status.value if is_coded else self._status_to_string(status),
+            "status": status.value,
```
mjnovice: Might want to test if reporting for non-legacy evals are not broken.

Sadly we don't have `reporting` being tested here as that requires a live project created on the alpha endpoint. 

And would be great if we can think of adding a test specifically for `reporting` probably by mocking the `Agents` backend endpoint, or maybe something else ?

src/uipath/_cli/_evals/_progress_reporter.py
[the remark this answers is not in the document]
Chibionos: this was a wrong assumption in the code @mjnovice i checked the backend  this is the right way for it. Let me update the tests for this case.

src/uipath/_cli/_evals/_progress_reporter.py
Chibionos: Tested it and it worked from python, we can test in alpha as the cut was already made.

src/uipath/_cli/_evals/_progress_reporter.py
```diff
@@ -110,17 +110,24 @@ def _format_error_message(self, error: Exception, context: str) -> None:
         self._rich_console.print(f"    • \u26a0  [dim]{context}: {error}[/dim]")
 
     def _status_to_string(self, status: EvaluationStatus) -> str:
```
mjnovice: are we even using this function ?

Chibionos: Correct this is not needed anymore as only integer is sent back

src/uipath/_cli/_evals/_progress_reporter.py
```diff
@@ -800,25 +807,38 @@ def _create_eval_run_spec(
             "inputs": eval_item.inputs,
         }
 
-        # For new coded evaluators (EvaluationItem), use evaluationCriterias
-        # For legacy evaluators (LegacyEvaluationItem), use expectedOutput
-        if isinstance(eval_item, EvaluationItem):
+        # For coded evaluators, use evaluationCriterias directly
```
smflorentino: seems worth unit testing?

PR #1029, opened 2025-12-17
feat: sanitize breakpoint names [PC-3577]

ionut-mihalache-uipath:
Breakpoints for low code agents don't stop the execution when the tool name has special characters. Sanitizing the breakpoint names will cause breakpoints to hit. After this PR is merged, I will do another PR for using in uipath-langchain-python the sanitize_name from uipath-python and remove sanitize_tool_name from uipath-langchain-python.

### DIFF
--- a/src/uipath/_cli/_debug/_bridge.py
+++ b/src/uipath/_cli/_debug/_bridge.py
@@ -23,6 +23,7 @@ from uipath.runtime.events import UiPathRuntimeStateEvent
 from uipath.runtime.resumable import UiPathResumeTriggerType
 
 from uipath._cli._utils._common import serialize_object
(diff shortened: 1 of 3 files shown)


src/uipath/_cli/_debug/_bridge.py
```diff
@@ -768,8 +769,9 @@ def _add_breakpoints(self, breakpoints: list[dict[str, Any]]) -> None:
                 else None
             )
             if node_name:
-                self.state.add_breakpoint(node_name)
-                logger.info(f"Breakpoint added: {node_name}")
+                sanitized_name = sanitize_string(node_name)
```
cristipufu: shouldn't we also do it for the Console Debugger?

PR #1048, opened 2025-12-23
feat: support 'same-as-agent' model option for legacy evaluators

Chibionos:
## Summary
- Add support for the `same-as-agent` model configuration in legacy LLM-based evaluators
- When an evaluator specifies `same-as-agent` as its model, it now resolves to the actual model from the agent runtime
- Added `LLMAgentRuntimeProtocol` for runtimes to provide model information

## Changes
- Added `LLMAgentRuntimeProtocol` - protocol for runtimes that can provide agent model info
- Updated `_get_agent_model()` to be async and query model from runtime (not factory)
- Added `_find_agent_model_in_runtime()` for recursive delegate traversal through runtime wrappers
- Updated `_load_evaluators()` to be async to support async model query
- Removed agent.json file fallback (runtime handles loading agent config)

## Architecture
The model resolution follows the runtime wrapper chain:
```
UiPathResumableRuntime (delegate property)
  └─> TelemetryRuntimeWrapper (get_agent_model delegates to _delegate)
        └─> AgentsLangGraphRuntime (implements LLMAgentRuntimeProtocol.get_agent_model)
```

## Companion PR
This PR works with [uipath-agents-python PR](https://github.com/UiPath/uipath-agents-python) that implements:
- `AgentsLangGraphRuntime.get_agent_model()` - loads model from agent.json
- `TelemetryRuntimeWrapper.get_agent_model()` - delegates to underlying runtime
- `AgentsRuntimeFactory` - removed get_agent_model (it's a runtime concept, not factory)

## Test plan
- [x] Unit tests for `LLMAgentRuntimeProtocol` integration
- [x] Tested with `calculator_same_as_agent` example containing evaluators with `"model": "same-as-agent"`
- [x] Verified both LLM-as-judge and Trajectory evaluators resolve model correctly
- [x] Verified recursive delegate traversal through runtime wrapper chain

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.3.0"
+version = "2.3.1"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/uv.lock
+++ b/uv.lock
@@ -2477,7 +2477,7 @@ wheels = [
 
 [[package]]
 name = "uipath"
-version = "2.3.0"
+version = "2.3.1"
 source = { editable = "." }
 dependencies = [
     { name = "click" },


akshaylive (commented):
IMO it's better to do something like this:

```
from typing import Protocol, runtime_checkable


@runtime_checkable
class LLMAgentProtocol(Protocol):
    def get_agent_model(self) -> str:
        ...
```

And switch the implementation of `_get_agent_model` to:
```
def _get_agent_model(self, runtime: UiPathRuntimeProtocol) -> str | None:
    if isinstance(runtime, LLMAgentProtocol):
        return runtime.get_agent_model()
    else:
        return None
```

That way, react agent can implement that method and it should work seamlessly.

src/uipath/_cli/_evals/_evaluator_factory.py
```diff
@@ -106,12 +109,15 @@ def _prepare_evaluator_config(data: dict[str, Any]) -> dict[str, Any]:
 
     @classmethod
     def create_evaluator(
-        cls, data: dict[str, Any], evaluators_dir: Path | None = None
+        cls,
+        data: dict[str, Any],
+        evaluators_dir: Path | None = None,
+        agent_model: str | None = None,
```
mjnovice: nit: I would suggest passing in `agent_model_settings` object instead, so that @mathurk @AAgnihotry do not have to do double work :)

src/uipath/_cli/_evals/_runtime.py
```diff
@@ -601,7 +631,58 @@ async def run_evaluator(
...
+            runtime: The runtime to check (may be a wrapper)
+
+        Returns:
+            The model name if found, None otherwise.
+        """
+        # Check if this runtime implements the protocol
+        if isinstance(runtime, LLMAgentRuntimeProtocol):
+            return runtime.get_agent_model()
+
+        # Check for delegate property (used by UiPathResumableRuntime, TelemetryRuntimeWrapper)
```
akshaylive: I'd prefer adding this method in UiPathResumableRuntime and TelemetryRuntimeWrapper but we can do this later..

PR #1053, opened 2025-12-27
feat: add model settings id parameter

mathurk:
## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.3.5.dev1010533572",

  # Any version from PR
  "uipath>=2.3.5.dev1010530000,<2.3.5.dev1010540000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath>=2.3.5.dev1010530000,<2.3.5.dev1010540000",
]
```

### DIFF
--- a/src/uipath/_cli/_evals/_runtime.py
+++ b/src/uipath/_cli/_evals/_runtime.py
@@ -292,7 +292,7 @@ class UiPathEvalRuntime:
     async def execute(self) -> UiPathRuntimeResult:
         # Configure model settings override before creating runtime
         await self._configure_model_settings_override()
-        
+
         runtime = await self.factory.new_runtime(
             entrypoint=self.context.entrypoint or "",
             runtime_id=self.execution_id,


src/uipath/_cli/_evals/_runtime.py
```diff
@@ -513,11 +519,97 @@ def _get_and_clear_execution_data(
...
+        target_model_settings = next(
+            (ms for ms in evaluation_set.model_settings if ms.id == self.context.model_settings_id),
+            None
+        )
+
+        if not target_model_settings:
+            logger.warning(f"Model settings ID '{self.context.model_settings_id}' not found in evaluation set")
+            return None
+
+        console.info(f"Found model settings: model='{target_model_settings.model_name}', temperature='{target_model_settings.temperature}'")
```
cristipufu: Please use logger, console is a CLI concept (imagine we extract the evals engine to another library)

src/uipath/_cli/_evals/_runtime.py
```diff
@@ -513,11 +519,97 @@ def _get_and_clear_execution_data(
...
+
+        # Early exit: if both values are "same-as-agent", no override needed
+        if (target_model_settings.model_name == "same-as-agent" and
+            target_model_settings.temperature == "same-as-agent"):
+            console.info("Both model and temperature are 'same-as-agent', no override needed")
+            return None
+
+        # Load the original entrypoint file
+        entrypoint_path = Path(self.context.entrypoint or "agent.json")
```
cristipufu: We shouldn't leak `agent.json` concepts here .. similar comment: https://github.com/UiPath/uipath-python/pull/1048#pullrequestreview-3606793664

akshaylive: +1. 

The other PR uses protocols to break this. Maybe we need to update the protocol method to return `EvaluationSetModelSettings` instead of model_name.

src/uipath/_cli/_evals/_models/_evaluation_set.py
```diff
@@ -78,6 +78,14 @@ class ModelSettings(BaseModel):
     max_tokens: int | None = Field(default=None, alias="maxTokens")
 
 
+class EvaluationSetModelSettings(BaseModel):
```
akshaylive: Should we reuse `ModelSettings` instead? `EvaluationSetModelSettings` seems to be a subset of `ModelSettings`, so it should be forward-compatible.

src/uipath/_cli/_evals/_runtime.py
```diff
@@ -550,12 +557,61 @@ def _get_and_clear_execution_data(
...
     async def execute_runtime(
         self,
         eval_item: EvaluationItem,
         execution_id: str,
         runtime: UiPathRuntimeProtocol,
     ) -> UiPathEvalRunExecutionOutput:
+        # Apply model settings override if specified
+        await self._configure_model_settings_override()
+
+        runtime = await self.factory.new_runtime(
```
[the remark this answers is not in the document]
mathurk: Looking at it again, I can override the runtime before creation, which would be better. I've gone ahead and made the fix, which removes this extra runtime creation

PR #1083, opened 2026-01-09
feat: add suspend/resume support for RPA invocations in evaluations

Chibionos:
## Summary
Adds support for suspend/resume pattern in evaluation runtime, enabling agents to suspend execution when invoking RPA processes and resume after completion.

## Key Changes

**Suspend Detection (_runtime.py)**
- Runtime detects `UiPathRuntimeStatus.SUSPENDED` after agent execution
- Extracts trigger information from suspended executions
- Skips evaluators during suspension (evaluators only run on SUCCESSFUL completion)
- Returns triggers in partial results for resume context

**Resume Support (cli_eval.py)**
- Added `--resume` flag to evaluation CLI
- Loads checkpoint state and continues from interruption point
- Passes resume context to runtime

**Trigger Pass-Through**
- Suspended executions return triggers with structure:
  - `interruptId`: Unique interrupt identifier
  - `triggerType`: Type of trigger (e.g., "Api")
  - `payload`: InvokeProcess details (process name, input arguments, folder path)

**Comprehensive Logging**
- Suspension detection events (🔴 DETECTED SUSPENSION)
- Trigger extraction details (📋 Extracted N trigger(s))
- Evaluator skip reasons (⏭️ Skipping evaluators)
- Trigger pass-through confirmation (✅ Passing through triggers)
- Resume mode indication (🟢 RESUMED execution)

## Testing
Testing is done via the sample agent in the companion PR: UiPath/uipath-langchain-python#397

## Related
Companion PR for sample agent: UiPath/uipath-langchain-python#397


mjnovice:
Just to understand, so the `suspend` and `resume` is actually being done by the `agent` and not the `orchestrator`, right ?

PR #1084, opened 2026-01-09
fix(SupportTargetOutputKey): support target output key

akshaylive:
Legacy exact match evaluator must support target output key.


mjnovice (approved):
LGTM, can we also add atleast one E2E test case for legacy eval with output_key  https://github.com/UiPath/uipath-python/blob/main/testcases/calculator-evals/run.sh#L11 ?

PR #1125, opened 2026-01-15
feat: support just latest version of guardrails validate api response

apetraru-uipath:
# Support Latest Version of Guardrails Validate API Response

## Summary
Updated `GuardrailsService.evaluate_guardrail()` to only support the latest API response format, removing backwards compatibility with legacy formats.

## Changes
- **Simplified response parsing**: Now only processes the `result` field from the latest API response format
- **Removed legacy format support**: No longer handles the legacy `validation_passed` field or `skip` field from older API versions

## Benefits
- **Cleaner implementation**: Reduced code complexity by removing backwards compatibility logic
- **Aligned with latest API**: Ensures the service matches the current guardrails validate API specification

### DIFF
--- a/src/uipath/platform/guardrails/_guardrails_service.py
+++ b/src/uipath/platform/guardrails/_guardrails_service.py
@@ -30,6 +30,32 @@ class GuardrailsService(BaseService):
     ) -> None:
         super().__init__(config=config, execution_context=execution_context)
 
+    @staticmethod
+    def _parse_result(result_str: str) -> GuardrailValidationResultType:
+        """Parse result string from API response to GuardrailValidationResultType.
+
+        Args:
+            result_str: The result string from the API response (e.g., "VALIDATION_FAILED").
+
+        Returns:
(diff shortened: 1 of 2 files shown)


src/uipath/platform/guardrails/_guardrails_service.py
```diff
@@ -66,39 +92,15 @@ def evaluate_guardrail(
...
-                else GuardrailValidationResultType.VALIDATION_FAILED
-            )
-
-        # Ensure result is always set (defensive check)
-        if result is None:
-            result = GuardrailValidationResultType.VALIDATION_FAILED
-
-        # Prepare model data with only the fields needed by GuardrailValidationResult
-        # (result and reason; ignore old fields like details, validation_passed, skip)
+        result = self._parse_result(response_data.get("result", ""))
```
valentinabojan: Why do we need to do manual parsing here? 
Is this not working? GuardrailValidationResult.model_validate(response.json())

PR #1137, opened 2026-01-16
fix: accept both aliases and field names when validating attachment

cristian-groza:
- Enable populate_by_name=True for the Attachment Pydantic model so it can be instantiated/validated using either field names (id, full_name, mime_type) or their aliases (ID, FullName, MimeType).

- Some callers provide attachment payloads using Python-style field names instead of the alias names, which caused validation to fail. This change makes the model more flexible and prevents unnecessary input transformations.

### DIFF
--- a/tests/sdk/services/test_attachments_service.py
+++ b/tests/sdk/services/test_attachments_service.py
@@ -915,7 +915,7 @@ class TestAttachmentsService:
         # Arrange
         attachment_key = uuid.UUID("12345678-1234-1234-1234-123456789012")
         file_content = b"Test file content for reading"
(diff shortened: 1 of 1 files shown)


src/uipath/platform/attachments/attachments.py
```diff
@@ -23,6 +23,7 @@ class Attachment(BaseModel):
     mime_type: str = Field(..., alias="MimeType")
     model_config = {
         "title": "UiPathAttachment",
+        "populate_by_name": True
```
cristipufu: populate by name is deprecated

PR #1148, opened 2026-01-18
fix: decouple mocking from evals

cristipufu:
## Description

Decouples mocking from evaluation context.

- Separates the mocking infrastructure from the evaluation system, enabling mocking to be used independently. 
- Mocking can now be configured without requiring a full evaluation item.

Ref: https://github.com/UiPath/uipath-python/pull/1117

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.5.27.dev1011484088",

  # Any version from PR
  "uipath>=2.5.27.dev1011480000,<2.5.27.dev1011490000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath>=2.5.27.dev1011480000,<2.5.27.dev1011490000",
]
```


src/uipath/_cli/_evals/mocks/llm_mocker.py
```diff
@@ -155,16 +156,16 @@ async def response(
                     },
                     "agentInfo": {  # This is incomplete
```
cristipufu: @akshaylive why do we need this `agentInfo`?

src/uipath/_cli/_evals/mocks/mocks.py
```diff
@@ -4,15 +4,15 @@
...
 from uipath._cli._evals.mocks.cache_manager import CacheManager
 from uipath._cli._evals.mocks.mocker import Mocker, UiPathNoMockFoundError
 from uipath._cli._evals.mocks.mocker_factory import MockerFactory
+from uipath._cli._evals.mocks.strategy import MockingStrategy
 
 # Context variables for evaluation items and mockers
-evaluation_context: ContextVar[EvaluationItem | None] = ContextVar(
-    "evaluation", default=None
+mocking_strategy_context: ContextVar[MockingStrategy | None] = ContextVar(
+    "mocking_strategy", default=None
```
cristipufu: do we actually need this, or can we rely on `mocker_context`?

akshaylive: Yes. The reason is because the mocking strategy can be stateful. For example, a mockito strategy could say "when input is any, return "foo" for the first invocation and "bar" for the second invocation". Without keeping the reference to MockingStrategy in the context, this is hard.

src/uipath/_cli/_evals/_runtime.py
```diff
@@ -868,8 +870,16 @@ async def _generate_input_for_eval(
         self, eval_item: EvaluationItem, runtime: UiPathRuntimeProtocol
     ) -> EvaluationItem:
         """Use LLM to generate a mock input for an evaluation item."""
+        expected_output = (
```
akshaylive: It's strange for input mocker to be using expectation values. The prompts were reused so we didn't think much of this. Do you know what happens in prod during simulations -- is it `{}`?

src/uipath/_cli/_evals/mocks/llm_mocker.py
akshaylive: The prompts were mostly copied over from URT (check L41). It does make sense for mocker to know about the agent's context -- especially the eval inputs. Maybe we need to create a separate class like this?:

```
class MockItem(BaseModel): # Equivalent to evaluation item
    inputs: Any
    name: str = Field(default="debug", ..)
    mocking_strategy: MockingStrategy
```

src/uipath/_cli/_evals/mocks/input_mocker.py
```diff
@@ -54,8 +56,10 @@ def get_input_mocking_prompt(
 
 @traced(name="__mocker__", recording=False)
 async def generate_llm_input(
-    evaluation_item: EvaluationItem,
+    mocking_strategy: InputMockingStrategy,
     input_schema: dict[str, Any],
+    expected_behavior: str,
```
akshaylive: Let's make this `Optional[str]` and `Optional[dict[str, Any]]` respectively please

PR #1152, opened 2026-01-19
feat: add discovery call to the llm gateway service

ionut-mihalache-uipath:
Summary

- Add fetch_llm_discovery function to the LLM Gateway service for fetching available models from the discovery endpoint

Changes

- Added fetch_llm_discovery(agenthub_config: str) function to _llm_gateway_service.py
- Exported the function from uipath.platform.chat module

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.5.11"
+version = "2.5.12"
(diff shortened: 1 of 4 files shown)


radu-mocanu (changes requested):
https://github.com/UiPath/uipath-python/pull/1152#discussion_r2705202932 
this is actually a good idea given we have multiple use cases for ahub. 
@ionut-mihalache-uipath let s wait a bit with this PR if possible. I'll expose AgentHub as a platform service and we can rebase on top of it

PR #1161, opened 2026-01-20
feat: add task recipient for guardrails

dianapirvulescu:
Add support to assign a HITL task to a user or a group. 
Added `TaskRecipient` object that holds information about the recipient of a task. At the moment both `assignee` and `task_recipient` can be used to assign a task but in the future only `task_recipient`  will remain and `assignee` will be removed,

### DIFF
--- a/src/uipath/platform/action_center/_tasks_service.py
+++ b/src/uipath/platform/action_center/_tasks_service.py
@@ -315,7 +315,7 @@ class TasksService(FolderContext, BaseService):
             headers=spec.headers,
         )
         json_response = response.json()
-        if assignee:
+        if assignee or recipient:
             spec = await _assign_task_spec(
                 self, json_response["id"], assignee, recipient
             )


src/uipath/platform/action_center/_tasks_service.py
```diff
@@ -113,13 +120,96 @@ def _retrieve_action_spec(
...
+                user_spec.method,
+                user_spec.endpoint,
+                json=user_spec.json,
+                content=user_spec.content,
+                headers=user_spec.headers,
+                scoped="org",
+            )
+            recipient_value = user_response.json().get("email")
+        if task_recipient.type == TaskRecipientType.GROUP_ID:
+            group_spec = _resolve_group(assignee)
```
ctiliescuuipath: this should be task_recipient.value instead of assignee

src/uipath/platform/action_center/tasks.py
```diff
@@ -15,6 +15,27 @@ class TaskStatus(enum.IntEnum):
     COMPLETED = 2
 
 
+class TaskRecipientType(str, enum.Enum):
+    """Task recipient type enumeration."""
+
+    USER_ID = "UserId"
+    GROUP_ID = "GroupId"
+    USER_EMAIL = "UserEmail"
```
valentinabojan: Let's call this simply EMAIL

PR #1178, opened 2026-01-22
feat: allow resource overwrites for mcp servers

edis-uipath:
Add the possibility to have resource overwrites on mcp servers through the retrieve action.


src/uipath/platform/orchestrator/_mcp_service.py
radu-mocanu: let s also apply the decorator here

PR #1182, opened 2026-01-22
feat: filter HTTP instrumentation spans from evaluation traces

AAgnihotry:
Implements whitelist filtering to remove POST/GET HTTP spans from evaluation traces, matching the fix applied in uipath-agents-python PRs #129 and #141.

Changes:
- Add uipath.custom_instrumentation=True to all evaluation runtime spans:
  - Evaluation Set Run
  - Evaluation
  - Evaluator
  - Evaluation output
- Update LlmOpsHttpExporter._should_drop_span() to use whitelist filtering
- Only export spans with uipath.custom_instrumentation=True
- Automatically filter out HTTP instrumentation spans (POST/GET)
- Preserve telemetry.filter="drop" for backward compatibility

This creates cleaner evaluation traces by removing auto-instrumentation noise while keeping only custom UiPath evaluation spans.

## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.5.33.dev1011824178",

  # Any version from PR
  "uipath>=2.5.33.dev1011820000,<2.5.33.dev1011830000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath>=2.5.33.dev1011820000,<2.5.33.dev1011830000",
]
```

Screenshot

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.5.32"
+version = "2.5.33"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/uv.lock
+++ b/uv.lock
@@ -2491,7 +2491,7 @@ wheels = [
 
 [[package]]
 name = "uipath"
-version = "2.5.32"
+version = "2.5.33"
 source = { editable = "." }
 dependencies = [
     { name = "applicationinsights" },


src/uipath/_cli/_evals/_runtime.py
```diff
@@ -878,11 +880,15 @@ async def execute_runtime(
...
+            # Don't pass execution_id to UiPathExecutionRuntime to avoid creating
+            # an extra "root" span. The parent "Evaluation" span already provides
+            # the necessary tracing context, and having the execution.id attribute set.
+            # We still pass the log_handler to capture logs.
             execution_runtime = UiPathExecutionRuntime(
                 delegate=eval_runtime,
                 trace_manager=self.trace_manager,
                 log_handler=log_handler,
-                execution_id=execution_id,
+                execution_id=None,
```
Chibionos: we need the execution_Id to the set here right ?

AAgnihotry: check the comment

AAgnihotry: Don't pass execution_id to UiPathExecutionRuntime to avoid creating
            # an extra "root" span. The parent "Evaluation" span already provides
            # the necessary tracing context, and having the execution.id attribute set.
            # We still pass the log_handler to capture logs.

PR #1274, opened 2026-02-06
feat: add app_name and recipient to WaitTask model

dianagrecu-uipath:
Add app_name and recipient to WaitTask model

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.8.0"
+version = "2.8.1"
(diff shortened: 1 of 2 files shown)


src/uipath/platform/common/interrupt_models.py
```diff
@@ -75,6 +75,8 @@ class WaitTask(BaseModel):
     action: Task
     app_folder_path: str | None = None
     app_folder_key: str | None = None
+    app_name: str | None = None
+    recipient: TaskRecipient | None = None
```
cotovanu-cristian: I think this model is only used when a task was already started (with a recipient on it)

cotovanu-cristian (approved):

PR #1315, opened 2026-02-14
feat: add local-only feature flags registry

cotovanu-cristian:
Introduce a simple feature flags config module that allows flags to be set programmatically via configure() and overridden locally with UIPATH_FEATURE_<name> environment variables.


cotovanu-cristian:
Closing this as I added it in `uipath-core-python` https://github.com/UiPath/uipath-core-python/pull/39

PR #1349, opened 2026-02-19
fix: handle exceptions in serialize_object

andreitava-uip:
Fixes payload serialization for SignalR events:
image


cristipufu (approved):

cristipufu:
@andreitava-uip we should update the serialize_defaults() method in uipath-core as well https://github.com/UiPath/uipath-core-python/blob/main/src/uipath/core/serialization/json.py

PR #1350, opened 2026-02-19
fix: save trace as agent run suspends and resumes

mathurk:
**The Problem**                                                                                                                                                                                   
                                                                                                                                                                                                
  When running evaluations locally on agents that use sub-agent tools (like "Backwards-String-Generator"), the evaluation process involves a suspend/resume cycle:                              
                                                                                                                                                                                                
  1. Initial run: The agent calls a sub-agent tool, which creates a remote job. 
  2. Resume run: A second CLI invocation picks up where the first left off, gets the job result, and completes the agent execution.

  The trajectory evaluator (which grades how well the agent followed expected steps by examining the trace) was scoring 0 for these agents, even though the agent was running as expected.

  ---
  **Root Cause 1: Spans lost across process boundary**

  OpenTelemetry spans are stored in memory by the ExecutionSpanExporter. When the first process suspends and exits, all
  32+ spans from that run are lost due to the line `self.span_exporter.clear(execution_id)`. The second process starts fresh with zero spans. The trajectory evaluator only sees the resume-phase spans and has no record of the agent's actual work.

  Fix: We added span persistence to SQLite. On suspend, all collected spans are serialized to JSON and saved to the existing __uipath/state.db database (which was already used for storing resume triggers). On resume, the saved spans are loaded back and prepended to the new spans. This required writing _serialize_span() and _deserialize_span() helper functions to convert OpenTelemetry ReadableSpan objects to/from JSON-compatible dicts.

  Result: Trajectory evaluator went from 0 → 50.

  ---
  **Root Cause 2: Resume-phase spans invisible to evaluator (exec_id=None)**

  Even with span persistence working, the trajectory evaluator scored 50 instead of 100. It could see the tool call (from saved first-run spans) but not the successful tool result (from resume-phase spans).

  Every span needs an execution.id attribute to be collected by the ExecutionSpanExporter. This ID is propagated from parent spans to child spans by UiPathExecutionTraceProcessorMixin.on_start() in the tracing infrastructure. However, this propagation requires parent_span.is_recording() to return True.

  On resume, the "Evaluation" span is restored as a NonRecordingSpan. NonRecordingSpan.is_recording() returns False, so execution.id propagation breaks at this boundary. Resume-phase spans never get execution.id, so the exporter silently drops them.

  The tracing infrastructure code (uipath.core.tracing.processors) is in a separate installed package we can't modify. But the eval-specific ExecutionSpanProcessor (which extends it) is in our code.

  Fix: Added a fallback in ExecutionSpanProcessor.on_start(): after the parent's propagation attempt, if execution.id is still missing, read it from the execution_id_context ContextVar (which is already set before the runtime executes). This ensures all spans during eval execution get tagged correctly, regardless of whether the parent span is recording.

  Result: Trajectory evaluator went from 50 → 100.
  
<img width="1252" height="642" alt="Screenshot 2026-02-19 at 4 24 45 PM" src="https://github.com/user-attachments/assets/373a4319-a211-4b8d-8657-ad34a70b1969" />


<!-- DEV_PACKAGE_START -->
## Development Package

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.8.47.dev1013504954",

  # Any version from PR
  "uipath>=2.8.47.dev1013500000,<2.8.47.dev1013510000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = [
    "uipath>=2.8.47.dev1013500000,<2.8.47.dev1013510000",
]
```
<!-- DEV_PACKAGE_END -->

### DIFF
--- a/src/uipath/_cli/_evals/_runtime.py
+++ b/src/uipath/_cli/_evals/_runtime.py
@@ -85,7 +85,10 @@ from ._models._output import (
     convert_eval_execution_output_to_serializable,
 )
 from ._span_collection import ExecutionSpanCollector
-from ._span_persistence_helpers import deserialize_span, serialize_span
+from ._span_persistence_helpers import (
+    load_execution_spans,
+    save_execution_spans,
+)
 from .mocks.mocks import (
     cache_manager_context,
     clear_execution_context,
@@ -869,11 +872,7 @@ class UiPathEvalRuntime:
                 # 4. Pass this map to the delegate runtime
                 if self.context.resume:
                     logger.info(f"Resuming evaluation {eval_item.id}")
-                    input = (
-                        (input_overrides or None)
-                        if self.context.job_id is None
-                        else None
-                    )
+                    input = input_overrides if self.context.job_id is None else None
                 else:
                     input = inputs_with_overrides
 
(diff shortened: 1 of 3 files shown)


Chibionos (changes requested):
need tests

Chibionos (approved):

PR #1391, opened 2026-02-28
feat: add WorkloadExecuting status to evaluation runs

mjnovice:
## Summary
- Adds `WorkloadExecuting` status to distinguish agent execution from evaluation/scoring in the eval run lifecycle
- Eval runs are now created with `WorkloadExecuting` (4) status instead of `IN_PROGRESS` (1)
- After agent execution completes, a new `AGENT_EXECUTION_COMPLETED` event transitions the run to `IN_PROGRESS` before evaluators execute
- New flow: `WorkloadExecuting → Running → Completed/Failed`

Depends on https://github.com/UiPath/Agents/pull/4529   

## Changes
- `events.py`: Add `AGENT_EXECUTION_COMPLETED` event type and `AgentExecutionCompletedEvent` model
- `runtime.py`: Publish `AGENT_EXECUTION_COMPLETED` after agent finishes, before evaluators run
- `_progress_reporter.py`: Add `WORKLOAD_EXECUTING=4` to enum, create eval runs with new status, handle event to transition to `IN_PROGRESS`

## Related
- Backend PR: https://github.com/UiPath/Agents/pull/4529

## Test plan
- [ ] Run evaluations end-to-end and verify eval run starts in `WorkloadExecuting` status
- [ ] Verify status transitions to `Running` after agent execution completes
- [ ] Verify status transitions to `Completed`/`Failed` after evaluators finish
- [ ] Verify resume mode still works correctly
- [ ] Run existing eval tests (`pytest tests/cli/eval/`)

### DIFF
--- a/packages/uipath/src/uipath/_cli/_evals/_progress_reporter.py
+++ b/packages/uipath/src/uipath/_cli/_evals/_progress_reporter.py
@@ -31,6 +31,7 @@ from uipath.eval.models import EvalItemResult, ScoreType
 from uipath.eval.models.evaluation_set import EvaluationItem
 from uipath.eval.runtime.events import (
     EvalRunCreatedEvent,
+    EvalRunStatusUpdateEvent,
     EvalRunUpdatedEvent,
     EvalSetRunCreatedEvent,
     EvalSetRunUpdatedEvent,
@@ -47,6 +48,8 @@ class EvaluationStatus(IntEnum):
     IN_PROGRESS = 1
     COMPLETED = 2
     FAILED = 3
+    WORKLOAD_EXECUTING = 4
+    WORKLOAD_FAILED = 5
 
 
 class StudioWebProgressItem(BaseModel):
@@ -473,6 +476,7 @@ class StudioWebProgressReporter:
         evaluators: dict[str, BaseEvaluator[Any, Any, Any]],
         is_coded: bool = False,
         spans: list[Any] | None = None,
+        workload_failed: bool = False,
     ):
         """Update an evaluation run with results."""
         coded_evaluators: dict[str, BaseEvaluator[Any, Any, Any]] = {}
@@ -512,6 +516,7 @@ class StudioWebProgressReporter:
                 actual_output=sw_progress_item.agent_output,
                 success=sw_progress_item.success,
                 is_coded=is_coded,
+                workload_failed=workload_failed,
             )
         else:
             spec = self._update_eval_run_spec(
@@ -522,6 +527,7 @@ class StudioWebProgressReporter:
                 actual_output=sw_progress_item.agent_output,
                 success=sw_progress_item.success,
                 is_coded=is_coded,
+                workload_failed=workload_failed,
             )
 
         await self._client.request_async(
@@ -545,6 +551,24 @@ class StudioWebProgressReporter:
         spec = self._update_eval_set_run_spec(
             eval_set_run_id, evaluator_scores, is_coded, success
         )
+        await self._client.request_async(
+            method=spec.method,
+            url=spec.endpoint,
+            params=spec.params,
+            json=spec.json,
+            headers=spec.headers,
+            scoped="org" if self._is_localhost() else "tenant",
+        )
+
+    @gracefully_handle_errors
+    async def move_eval_run_status(
+        self,
+        eval_run_id: str,
+        status: EvaluationStatus,
+        is_coded: bool = False,
+    ):
+        """Update an evaluation run's status without changing results."""
+        spec = self._update_eval_run_status_spec(eval_run_id, status, is_coded)
         await self._client.request_async(
             method=spec.method,
             url=spec.endpoint,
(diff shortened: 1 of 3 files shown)


packages/uipath/src/uipath/eval/runtime/runtime.py
```diff
@@ -626,6 +628,17 @@ async def _execute_eval(
                     # The evaluation will be completed when resumed
                     return evaluation_run_results
 
+                # Agent execution succeeded - mark flag and move to Running
+                # (evaluation/scoring phase is about to begin)
+                agent_execution_succeeded = True
+                await self.event_bus.publish(
```
Chibionos, line 634: This `MOVE_EVAL_RUN_TO_RUNNING` is published with `wait_for_completion=False`, and the terminal `UPDATE_EVAL_RUN` later in the same method is also `wait_for_completion=False`. `EventBus.publish` schedules each handler via `asyncio.create_task` without awaiting, so the two are detached tasks. Both handlers issue a PUT mutating the same eval-run status (this one -> IN_PROGRESS, the later one -> COMPLETED/FAILED). There is no ordering guarantee between two independently scheduled tasks, so the IN_PROGRESS PUT can land *after* the terminal PUT, leaving the run stuck displaying "Running" even though it finished. Evaluator runtime between the two publishes usually lets MOVE win, but the ordering is not enforced. Consider publishing this event with `wait_for_completion=True` (or awaiting the move in-line before scoring) so the Running transition is durably ahead of the terminal update.

packages/uipath/src/uipath/eval/runtime/runtime.py
```diff
@@ -626,6 +628,17 @@ async def _execute_eval(
                     # The evaluation will be completed when resumed
                     return evaluation_run_results
 
+                # Agent execution succeeded - mark flag and move to Running
+                # (evaluation/scoring phase is about to begin)
+                agent_execution_succeeded = True
```
Chibionos, line 633: `agent_execution_succeeded` is set `True` here, just before the move publish, and the outer `except` reports `workload_failed=not agent_execution_succeeded`. This produces the right result today (an exception in `execute_runtime` keeps it False -> WORKLOAD_FAILED; an exception during scoring keeps it True -> FAILED), but the invariant is implicit and easy to break if this block is ever reordered. A one-line comment stating "True means the agent workload completed; later failures are scoring failures, not workload failures" — or tracking the phase explicitly — would make the contract robust to future edits.

packages/uipath/src/uipath/_cli/_evals/_progress_reporter.py
```diff
@@ -1187,6 +1261,33 @@ def _update_coded_eval_run_spec(
...
+        logger.debug(
+            f"Updating eval run status: evalRunId={eval_run_id}, status={status.name}"
+        )
+
+        return RequestSpec(
+            method="PUT",
+            endpoint=Endpoint(
+                f"{self._get_endpoint_prefix()}execution/agents/{self._project_id}/{endpoint_suffix}evalRun"
+            ),
+            json=payload,
```
Chibionos, line 1287: `test_progress_reporter.py` is unchanged, so none of the new behavior is covered: the create spec now emitting `WORKLOAD_EXECUTING` (status 4), the `WORKLOAD_FAILED` (status 5) branch in `_update_eval_run_spec` / `_update_coded_eval_run_spec`, and this new `_update_eval_run_status_spec`. The analogous COMPLETED/FAILED branches already have spec tests (`test_update_coded_eval_run_spec_with_failure`, `test_update_legacy_eval_run_spec_with_failure`), so this is a coverage gap rather than an untestable area. Please add: a create-spec test asserting `status == 4`, an update-spec test with `workload_failed=True` asserting `status == 5`, and a test of this status-only spec's method (PUT), endpoint, and `{evalRunId, status}` payload.

Chibionos (commented):
Focused, well-scoped change that adds a two-phase status lifecycle (WorkloadExecuting -> Running -> terminal) to evaluation runs. The new enum values, the workload_failed flag plumbing, and the status-only update spec are all consistent with the file's existing endpoint and payload conventions (slop meter clean, 0/0). Two things to address before merge, neither blocking: (1) the MOVE-to-Running event and the terminal status update are both fire-and-forget, so under adverse timing the run can stick in Running - consider making the Running transition ordered; (2) none of the new status transitions have unit tests, despite the analogous COMPLETED/FAILED branches already being covered. A Codex recheck flagged a project_id-vs-agent_id endpoint bug, but that was verified against the head blob as a false positive - every eval-run endpoint in this file uses project_id and no _agent_id attribute exists.

### ⚠️ Risks to validate
- MOVE-to-Running and the terminal COMPLETED/FAILED PUT are both fire-and-forget; ordering is unsynchronized.
- No tests added for any of the new status transitions.

_Please confirm each is handled, or explain why it does not apply._
---
🔍 Reviewed with `/deep-review:deep-review` + `/adversarial-pr-review` · human-approved via team-review
`skill: UiPath/skills-internal@e83f124` · `model: claude-opus-4-8 (1M context)` · `agent: Claude Code`

PR #1428, opened 2026-03-09
(wip) feat(tracing): derive LLM Ops URL source param from span Source field

saksharthakkar:
## Summary
- Add `SOURCE_NAMES` mapping (int → string) for SourceEnum values in `_span_utils.py`
- Update `LlmOpsHttpExporter._build_url()` to derive the `source` URL query parameter from the span's `Source` field instead of hardcoding `"Robots"`
- Default remains `"Robots"` (SourceEnum=4) when Source is missing or unknown
- Fixes mismatch where agents sent `Source: 1` (Agents) in span payload but `source=Robots` in URL

## Test plan
- [x] New test `test_build_url_uses_span_source` covers: Agents(1), Robots(4), Playground(7), missing Source, unknown int(99)
- [x] Existing `test_uipath_span_source_defaults_to_robots` passes
- [x] Existing `test_uipath_span_source_override_with_uipath_source` passes
- [x] Full otel_exporters test suite (1590 tests) passes with no regressions

image

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/common/_span_utils.py
+++ b/packages/uipath-platform/src/uipath/platform/common/_span_utils.py
@@ -18,6 +18,17 @@ logger = logging.getLogger(__name__)
 # SourceEnum.Robots = 4 (default for Python SDK / coded agents)
 DEFAULT_SOURCE = 4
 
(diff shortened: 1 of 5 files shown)


saksharthakkar:
creating new PR for this

PR #1542, opened 2026-04-02
fix(core): prefer OTEL trace ID over external span when trace IDs differ

cfauchere:
## Summary
When both an OTEL current span (agent runtime) and an external span (OpenInference LangChain instrumentor) exist with different trace IDs, `get_parent_context()` now returns the OTEL span.

## Problem
The agent runtime sets an OTEL span with the agent's trace ID. The OpenInference LangChain instrumentor creates its own spans with a separate trace ID. When the Python SDK makes HTTP calls to downstream services (ECS, LLM Gateway), `_get_bottom_most_span()` would pick the OpenInference span because the depth-based tiebreaker defaulted to the external span.

Result: the `x-uipath-traceparent-id` header carried a different trace ID than the agent trace, making it impossible to correlate agent traces with CG/ECS logs.

## Fix
In `_get_bottom_most_span()`, when the two spans have different trace IDs, return the OTEL current span immediately (before the depth comparison). The existing depth logic still applies when trace IDs match.

## Validation
Tested locally: ran an agent with a CG semantic search tool, confirmed the header's trace ID matches the agent's TraceId.

## Test plan
- [ ] Verify trace ID in `x-uipath-traceparent-id` header matches agent TraceId
- [ ] Verify existing span parent-child logic still works when trace IDs match
- [ ] Verify no regression when only one span exists (no external provider)

### uipath

```toml
[project]
dependencies = [
  # Exact version:
  "uipath==2.10.43.dev1015425951",

  # Any version from PR
  "uipath>=2.10.43.dev1015420000,<2.10.43.dev1015430000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### uipath-core

```toml
[project]
dependencies = [
  # Exact version:
  "uipath-core==0.5.11.dev1015425951",

  # Any version from PR
  "uipath-core>=0.5.11.dev1015420000,<0.5.11.dev1015430000"
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
uipath-core = { index = "testpypi" }
```

### DIFF
--- a/packages/uipath-core/tests/tracing/test_span_nesting.py
+++ b/packages/uipath-core/tests/tracing/test_span_nesting.py
@@ -616,7 +616,7 @@ def test_ctx_parameter_required_when_external_deeper_than_current(
         # the OTEL current span takes priority (agent trace ID preservation).
         # The non-recording span should be parented to otel_span, not external_deep.
         non_recording_parented_to_otel = False
-        for span_id, parent_id in _span_registry._parent_map.items():
+        for _span_id, parent_id in _span_registry._parent_map.items():
             if parent_id == otel_span_id:
                 non_recording_parented_to_otel = True
                 break


saksharthakkar:
@akshaylive I vaguely remember for simulated tools you had a scenario where you needed Openinference trace id and not manual instrumentation Id... please take a look, just wanted to check if your scenario is not affected by this change.

PR #1561, opened 2026-04-13
fix: pass through error details to CAS for unrecognized error IDs [JAR-9616]

andrewwan-uipath:
- https://uipath.atlassian.net/browse/JAR-9616
- Pass through actual error details to CAS for unrecognized error IDs instead of showing a generic "An unexpected error has occurred." message                                                                     
 - Known error IDs (licensing, max steps, incomplete response) continue to use hardcoded CAS messages

### DIFF
--- a/packages/uipath/pyproject.toml
+++ b/packages/uipath/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.10.48"
+version = "2.10.49"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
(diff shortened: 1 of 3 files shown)


maxduu:
Oops, my bad for missing this one for review!

To confirm, this essentially shows the same error as it would in Orchestrator, correct? 

I was doing some thinking and maybe we shouldn't actually show the full code error to end-users. Instead we might want to give a message like "Inspect Orchestrator job" on some surfaces with a link to the job that failed so that way admins can inspect it, but end-users (e.g. those on iframe chats, anonymous chats, etc.) won't be getting "too much info" leaked. We can give it some more thought but let's not merge this just yet and I can start a discussion regarding final experience!

andrewwan-uipath:
> Oops, my bad for missing this one for review!
> 
> To confirm, this essentially shows the same error as it would in Orchestrator, correct?
> 
> I was doing some thinking and maybe we shouldn't actually show the full code error to end-users. Instead we might want to give a message like "Inspect Orchestrator job" on some surfaces with a link to the job that failed so that way admins can inspect it, but end-users (e.g. those on iframe chats, anonymous chats, etc.) won't be getting "too much info" leaked. We can give it some more thought but let's not merge this just yet and I can start a discussion regarding final experience!

oh oops also forgot about this PR. Sounds good, we can revisit this, going to focus on Flow for now

PR #1591, opened 2026-04-24
feat: add @resource_override and folder_path support to MemoryService

mjnovice:
## Summary

- Apply `@resource_override(resource_type="memorySpace")` to `create`/`create_async` so folder bindings are automatically resolved from resource overwrites (matching assets, ECS indexes, etc.)
- Add `folder_path` parameter to all public methods as an alternative to `folder_key`
- Plumb `folder_path` through to private spec builders via the existing `_resolve_folder` method

### Why

The memory service was the only resource type that required callers to manually read `_resource_overwrites` to resolve folder bindings. Other services (context grounding, assets, connections) use the `@resource_override` decorator for automatic resolution. This was flagged in UiPath/uipath-agents-python#374 review.

### Changes

**`_memory_service.py`**:
- `create` / `create_async`: Added `@resource_override(resource_type="memorySpace")` decorator + `folder_path` param
- `list` / `list_async`: Added `folder_path` param
- `search` / `search_async`: Added `folder_path` param
- `escalation_search` / `escalation_search_async`: Added `folder_path` param
- `escalation_ingest` / `escalation_ingest_async`: Added `folder_path` param
- All private spec builders updated to accept and forward `folder_path`

### Notes

- `@resource_override` is only applied to `create`/`create_async` since they have a `name` parameter (the decorator matches on `name` by default). `list`/`search`/`escalation_*` methods use `memory_space_id` or OData filters, so the decorator doesn't apply.
- The `memorySpace` resource type was already registered in `GenericResourceOverwrite`.
- Backwards compatible — `folder_path` defaults to `None`.

## Test plan

- [ ] Verify `create` with resource overwrites resolves folder correctly
- [ ] Verify `list`/`search` with `folder_path` param works

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/memory/_memory_service.py
+++ b/packages/uipath-platform/src/uipath/platform/memory/_memory_service.py
@@ -10,6 +10,7 @@ from typing import Any, Optional
 from uipath.core.tracing import traced
 
 from ..common._base_service import BaseService
+from ..common._bindings import resource_override
 from ..common._config import UiPathApiConfig
 from ..common._execution_context import UiPathExecutionContext
 from ..common._folder_context import FolderContext, header_folder
@@ -50,6 +51,7 @@ class MemoryService(FolderContext, BaseService):
 
     # ── Memory space operations (ECS) ──────────────────────────────────
 
+    @resource_override(resource_type="memorySpace")
     @traced(name="memory_create", run_type="uipath")
     def create(
         self,
@@ -57,6 +59,7 @@ class MemoryService(FolderContext, BaseService):
         description: Optional[str] = None,
         is_encrypted: Optional[bool] = None,
         folder_key: Optional[str] = None,
+        folder_path: Optional[str] = None,
     ) -> MemorySpace:
         """Create a new memory space.
 
@@ -65,11 +68,12 @@ class MemoryService(FolderContext, BaseService):
             description: Optional description (max 1024 chars).
             is_encrypted: Whether the memory space should be encrypted.
             folder_key: The folder key for the operation.
+            folder_path: The folder path for the operation.
 
         Returns:
             MemorySpace: The created memory space.
         """
-        spec = self._create_spec(name, description, is_encrypted, folder_key)
+        spec = self._create_spec(name, description, is_encrypted, folder_key, folder_path)
         response = self.request(
             spec.method,
             spec.endpoint,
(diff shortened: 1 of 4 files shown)


packages/uipath-platform/src/uipath/platform/memory/_memory_service.py
CalebMartinUiPath, line 222: is there a reason we don't add the resource override to the search methods?

CalebMartinUiPath (approved):

packages/uipath-platform/src/uipath/platform/memory/_memory_service.py
mjnovice: Search methods don't need the `@resource_override` decorator because they receive **already-resolved IDs** (`memory_space_id` as a GUID + `folder_key` as a GUID), not string-based resource names/paths.

The decorator is designed to intercept and override `name` + `folder_path` at the API entry point — which is relevant for `create` (where you pass a name like `"Due Diligence"` and a `folder_path` like `"solution_folder"`). The overwrite context maps `memorySpace.Due Diligence` → the real folder path.

For search/ingest, the caller has already resolved the memory space (via `list()` or from config) and has the folder_key. The Temporal backend follows the same pattern — it resolves the folder once in `ResolveMemorySpaceAsync()` and passes the resolved `memorySpace.FolderKey` GUID to both search and ingest calls (EscalationToolExecutor.cs lines 657-658, 519-520).

That said, we did add `folder_path` as a parameter to search/list methods so callers who only have a path (not a key) can pass it and let the SDK resolve it internally via `_resolve_folder()`.

mjnovice: *Updated (removing internal code references):*

Search methods don't need the `@resource_override` decorator because they receive **already-resolved IDs** (`memory_space_id` as a GUID + `folder_key` as a GUID), not string-based resource names/paths.

The decorator intercepts and overrides `name` + `folder_path` at the API entry point — relevant for `create` where you pass a name and folder path that may need overwrite resolution. For search/ingest, the caller already has resolved IDs.

The backend follows the same pattern — folder is resolved once upfront and the resolved GUID is reused for all subsequent memory operations (search, ingest).

We did add `folder_path` as a parameter to search/list methods so callers who only have a path (not a key) can pass it and let the SDK resolve it internally via `_resolve_folder()`.

PR #1594, opened 2026-04-28
feat: add is_case_manager field in agent declaration

dianagrecu-uipath:
Add is_case_manager field in agent declaration

### DIFF
--- a/packages/uipath/src/uipath/agent/models/agent.py
+++ b/packages/uipath/src/uipath/agent/models/agent.py
@@ -1156,6 +1156,7 @@ class AgentMetadata(BaseCfg):
     """Agent metadata model."""
 
     is_conversational: bool = Field(alias="isConversational")
+    is_case_manager: bool = Field(default=False, alias="isCaseManager")
     storage_version: str = Field(alias="storageVersion")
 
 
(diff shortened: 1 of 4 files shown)


packages/uipath/src/uipath/agent/models/agent.py
```diff
@@ -1216,6 +1217,15 @@ def is_conversational(self) -> bool:
                 return metadata.is_conversational
         return False
 
+    @property
+    def is_case_manager(self) -> bool:
+        """Checks if the agent is a case manager agent."""
+        if hasattr(self, "metadata") and self.metadata:
+            metadata = self.metadata
+            if hasattr(metadata, "is_case_manager"):
+                return metadata.is_case_manager
```
andreitava-uip: nit: I think this is redundant, if metadata is present then it should always have `is_case_manager` since it's not an optional field.

packages/uipath/tests/agent/models/test_agent.py
```diff
@@ -3283,6 +3283,119 @@ def test_is_conversational_false_by_default(self):
         assert config.is_conversational is False
 
 
+class TestAgentDefinitionIsCaseManager:
+    """Tests for AgentDefinition.is_case_manager property."""
+
+    def test_is_case_manager_true_when_metadata_set(self):
+        """Returns True when metadata.is_case_manager is True."""
+        json_data = {
```
andreitava-uip, line 3291: nit: could extract a helper for this and reduce duplication

andreitava-uip (approved):

packages/uipath/src/uipath/agent/models/agent.py
ionut-mihalache-uipath: in frontend it is an optional field.

andreiancuta-uipath: yes, but in the model it has a default. so only `metadata` needs to be checked

PR #1605, opened 2026-04-29
feat: add support for system indexes [ECS-1791]

gcuip:


packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
```diff
@@ -256,37 +257,107 @@ async def retrieve_across_folders_async(
             ContextGroundingIndex.model_validate(item) for item in response["value"]
         ]
 
+    @traced(name="contextgrounding_retrieve_system_indexes", run_type="uipath")
+    def retrieve_system_indexes(
```
radu-mocanu: would keep this as a private untraced method.

packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
```diff
@@ -1911,6 +2026,16 @@ def _ingest_spec(
             },
         )
 
+    @staticmethod
+    def _odata_name_filter(name: str) -> str:
```
radu-mocanu, line 2004: should we maybe make this generic and move it one level up (base class) so any downstream service can benefit from the encoding?

packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
gcuip: it will be useful for customers to see why they can run debug projects and where the index came from, but when running a production / deployed project, they'll get a different index (or an error if they have manually deleted it).

packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
gcuip: we can extract it later if it's needed by other services.

radu-mocanu (approved):

PR #1609, opened 2026-05-01
feat: add client side tools to bridge and new cas events

norman-le:
```
## uipath-python

- Added ClientSide tool type to agent model so the runtime can distinguish client-side tools from server tools
- Bridge emits executingToolCall event on durable interrupt when is_execution_phase marker is set, enabling the client to know when to execute the tool
- Bridge routes endToolCall from the client back to the agent, completing the interrupt/resume cycle

## uipath-langchain-python

- Client-side tool factory creates StructuredTool instances that use @durable_interrupt to pause the graph and wait for client execution
- @mockable wraps @durable_interrupt (as a separate outer function) so evals can simulate client-side tools without hitting GraphInterrupt
- CLIENT_SIDE_TOOL_MARKER defined in hitl.py as the single source of truth for identifying client-side tools across the codebase
- Error-handling wrapper (RunnableCallableWithTool) now preserves .tool reference for client-side tools, enabling runtime discovery
- Runtime discovers client-side tools from the compiled graph via shared _iter_graph_tools() iterator, used by both confirmation and client-side tool detection
- MessageMapper emits executingToolCall only for server tools without interrupts; client-side and confirmation tools get it from the bridge instead
- MessageMapper suppresses endToolCall for client-side tool results since the client already produced the result
- MessageMapper sets isClientSideTool and outputSchema on startToolCall events so the client knows to show the input form
- 9 new tests covering tool discovery, executingToolCall emission rules, and endToolCall suppression

## uipath-agents-python

- Graph builder strips CLIENT_SIDE_TOOL_MARKER during evals so @durable_interrupt is bypassed and @mockable can intercept

## AgentInterfaces

- CAS tracks is_client_side_tool on tool calls in the database and forwards client-originated endToolCall to the agent
- CAS validates client-side tool declarations against the agent definition and guards against unauthorized endToolCall for non-client-side tools
- React SDK listens for executingToolCall events and renders a form (ClientSideToolWidget) using the tool's output schema for the user to fill in
- ClientSideToolWidget reuses the existing ToolConfirmationFormContent component
- Added type schemas for executingToolCall events and client-side tool declarations
```

Videos in CAS localhost (Agent Builder changes already published, but this will be the debug experience as well when the sdk is upgraded):

Without tool confirmation for client side tool:

https://github.com/user-attachments/assets/48bc0979-0f51-471c-90aa-af561d876c31

With tool confirmation for client side tool:

https://github.com/user-attachments/assets/384ba36d-db2d-478e-8a74-8b3de0501a06

URT Eval:
image

Related to changes in other PRs:
CAS: https://github.com/UiPath/AgentInterfaces/pull/949
uipath-langchain-python (tool definition): https://github.com/UiPath/uipath-langchain-python/pull/819
uipath-python (bridge changes): https://github.com/UiPath/uipath-python/pull/1609
uipath-agents-python (skipping for evals, so it can be mocked): https://github.com/UiPath/uipath-agents-python/pull/485
Another Agents PR fixing some issues with debug: https://github.com/UiPath/Agents/pull/5250

### DIFF
--- a/packages/uipath/src/uipath/_cli/_chat/_bridge.py
+++ b/packages/uipath/src/uipath/_cli/_chat/_bridge.py
@@ -365,35 +365,12 @@ class SocketIOChatBridge:
             raise RuntimeError(f"Failed to send exchange error event: {e}") from e
 
     async def emit_interrupt_event(self, resume_trigger: UiPathResumeTrigger):
-        """No-op.
+        """Emit an executingToolCall event if the trigger is marked with is_execution_phase.
 
-        Tool confirmation — the only interrupt pattern CAS uses today — is
-        handled end-to-end via ``startToolCall`` with ``requireConfirmation:
-        true`` paired with ``wait_for_resume()``. This is deliberately
-        simpler than the old interrupt-based flow: CAS needs
-        ``requireConfirmation`` on the tool call event itself to render the
-        confirmation UI, so a parallel ``startInterrupt`` event would be
-        redundant.
-
-        The only hypothetical reason to put work here is a generic,
-        non-tool-call agent interrupt (e.g. a coded agent calling
-        ``interrupt("do you want to continue?")``). Nothing uses that today
-        and it's not a near-term requirement — the method is kept for
-        generic flexibility.
+        Called by the runtime loop for every durable interrupt. Emits executingToolCall
+        for triggers that signal execution is about to begin, The
+        is_execution_phase marker ensures it fires exactly once per tool call.
         """
-        return None
-
-    async def emit_executing_tool_call_event(
-        self, resume_trigger: UiPathResumeTrigger
-    ) -> None:
-        """Emit an executingToolCall event for client-side tool execution.
-
-        Only emits for triggers marked with is_execution_phase=True.
-        This fires exactly once per client-side tool call — for Path 3 (no confirm)
-        and for Path 4 (after confirmation, on the execution interrupt).
-        Confirmation-only interrupts (Paths 2/4 first interrupt) are skipped.
-        """
-
         request = (
             resume_trigger.api_resume.request if resume_trigger.api_resume else None
         )
@@ -408,9 +385,6 @@ class SocketIOChatBridge:
         tool_input = request.get("input")
 
         if not tool_call_id or not tool_name:
-            logger.info(
-                f"emit_executing_tool_call_event: missing tool_call_id or tool_name, skipping. tool_call_id={tool_call_id}, tool_name={tool_name}"
-            )
             return
 
         executing_event = UiPathConversationMessageEvent(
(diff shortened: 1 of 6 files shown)


packages/uipath/src/uipath/_cli/_chat/_bridge.py
```diff
@@ -378,6 +383,48 @@ async def emit_interrupt_event(self, resume_trigger: UiPathResumeTrigger):
...
         return None
 
+    async def emit_executing_tool_call_event(
+        self, resume_trigger: UiPathResumeTrigger
+    ) -> None:
+        """Emit an executingToolCall event for client-side tool execution.
+
+        Only emits for triggers marked with is_execution_phase=True.
+        This fires exactly once per client-side tool call — for Path 3 (no confirm)
+        and for Path 4 (after confirmation, on the execution interrupt).
```
JoshParkSJ: Maybe explain the paths or just reference the confluence doc https://uipath.atlassian.net/wiki/spaces/JAR/pages/90578682323/Client+side+tools+tool+confirmation+Sequence+Diagrams

packages/uipath/src/uipath/agent/models/agent.py
```diff
@@ -895,6 +890,15 @@ class AgentInternalToolResourceConfig(BaseAgentToolResourceConfig):
     )
 
 
+class AgentClientSideToolResourceConfig(BaseAgentToolResourceConfig):
+    """Resource config for client-side tools executed by the client SDK."""
+
+    type: Literal[AgentToolType.CLIENT_SIDE] = AgentToolType.CLIENT_SIDE
```
JoshParkSJ, line 896: Can we add an example maybe in agentinterfaces what an example client side tool would look like?

packages/uipath/src/uipath/_cli/_chat/_bridge.py
```diff
@@ -360,23 +365,39 @@ async def emit_exchange_error_event(self, error: Exception) -> None:
...
-        ``requireConfirmation`` on the tool call event itself to render the
-        confirmation UI, so a parallel ``startInterrupt`` event would be
-        redundant.
-
-        The only hypothetical reason to put work here is a generic,
-        non-tool-call agent interrupt (e.g. a coded agent calling
-        ``interrupt("do you want to continue?")``). Nothing uses that today
-        and it's not a near-term requirement — the method is kept for
-        generic flexibility.
+        """Emit an executingToolCall event if the trigger is marked with is_execution_phase.
```
maxduu, line 368: @JoshParkSJ Could you check this - I thought `emit_interrupt_event` should be kept a no-op? And maybe a separate function to emit conversational tool-call-executing events? Since we don't know if every `emit_interrupt_event` would be a tool-call, right

norman-le: I think it was a no-op since Josh moved it to use the requireConfirmation property instead, and Pufu didn't want to remove the emit interrupt event (for cases like this where we might need it).

I had another PR to modify the uipath-runtime-python but he told me to use the existing one as it was essentially emitting an interrupt on a trigger. But let me know if im misunderstanding that @JoshParkSJ

PR #1611, opened 2026-05-04
fix(connections): preserve filename and content type in multipart file uploads

Dennis-UiPath:
## Summary

`ConnectionsService._build_activity_request_spec` currently constructs
`files[key] = (key, val, None)` for every multipart parameter,
regardless of value type. Two consequences for any connector activity
that uploads a file:

1. **Filename is wrong.** `httpx` interprets the first element of a
   3-tuple as the filename for `Content-Disposition: form-data; name="…"; filename="…"`.
   Today every file is uploaded with the *form-field name* as its
   filename, e.g. `attachment[file]`. Downstream services that store
   attachments by name end up with literal `attachment_file_.` (Coupa
   sanitises `[`/`]`, then appends `.` because there's no extension).
2. **Content-Type is `None`.** httpx then defaults to `application/octet-stream`,
   so previewers / browsers cannot pick a sensible viewer.

The current source even flags the gap explicitly:

```python
# files not supported yet supported so this will likely not work
files[key] = (
    key,
    val,
    None,
)  # probably needs to extract content type from val since IS metadata doesn't provide it
```
[`packages/uipath-platform/src/uipath/platform/connections/_connections_service.py#L791-L796`](https://github.com/UiPath/uipath-python/blob/main/packages/uipath-platform/src/uipath/platform/connections/_connections_service.py#L791-L796)

A real-world manifestation: every Coupa attachment uploaded by an agent using `coupa.add_attachment` lands as `attachment_file_.` instead of the document's real filename.

## Fix

Branch on the value type to accept httpx's standard tuple shape, while keeping the existing bytes-only path working:

```python
if isinstance(val, tuple):
    # Caller supplied httpx's (filename, content[, content_type])
    # shape — pass through verbatim. This is the recommended path
    # for file uploads.
    files[key] = val
elif isinstance(val, (bytes, bytearray)) or hasattr(val, "read"):
    # Raw file content with no filename — fall back to the form-field
    # name (legacy behaviour). Backwards compatible with callers that
    # still pass bytes directly.
    files[key] = (key, val, "application/octet-stream")
else:
    # Scalar — send as a plain multipart form field, not a file part.
    # The (None, value) shape tells httpx to omit `filename=...` from
    # the Content-Disposition.
    files[key] = (None, str(val))
```

| Caller passes | Wire-level result |
|---|---|
| `("invoice.pdf", b"…", "application/pdf")` *(recommended)* | `Content-Disposition: form-data; name="attachment[file]"; filename="invoice.pdf"` + `Content-Type: application/pdf` |
| `("invoice.pdf", b"…")` *(2-tuple shorthand)* | filename preserved, httpx infers content type |
| `b"…"` *(legacy)* | unchanged behaviour: filename = form-field name |
| `"some string"` *(scalar)* | sent as plain form field — `filename=…` is **omitted** |

Backwards compatible: existing callers passing raw `bytes` keep working with no observable change on the wire.

## Tests

Added `TestMultipartFileUpload` in `packages/uipath-platform/tests/services/test_connections_service.py` with four cases that inspect the serialized multipart body:

- `test_invoke_activity_multipart_tuple_3_preserves_filename` — asserts `filename="invoice.pdf"` and `Content-Type: application/pdf` land in the wire body.
- `test_invoke_activity_multipart_tuple_2_preserves_filename` — 2-tuple shorthand preserves filename.
- `test_invoke_activity_multipart_bytes_backwards_compatible` — legacy raw-bytes callers keep behaviour: filename = form-field name, `application/octet-stream` content type.
- `test_invoke_activity_multipart_scalar_is_plain_form_field` — non-file params (e.g. `payload="{}"`) are emitted without a bogus `filename=` in `Content-Disposition`.

### Test results

Ran from `packages/uipath-platform/`:

- `uv run pytest tests/services/test_connections_service.py` → **57 passed** (53 pre-existing + 4 new).
- `uv run pytest tests/` → **1105 passed, 7 skipped** (the skipped ones are pre-existing LLM integration tests gated on real credentials).
- `uv run ruff check src/uipath/platform/connections/_connections_service.py` → clean.

### Verification that the new tests catch the bug

I `git stash`ed the fix and re-ran `pytest …::TestMultipartFileUpload` — 3 of the 4 new tests fail against the unpatched serializer (tuple_3, tuple_2, scalar). The bytes-backwards-compat test passes either way, because httpx defaults a `None` content type to `application/octet-stream` on the wire.

## End-to-end validation

The downstream change in our agent that motivated this — using the patched SDK to send the real document filename to Coupa via `add_attachment` — was verified end-to-end against the staging Coupa instance: same byte payload as before, but the attachment is now stored with the real filename (`invoice_<id>.pdf`) instead of `attachment_file_.`.

### DIFF
--- a/packages/uipath-platform/pyproject.toml
+++ b/packages/uipath-platform/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-platform"
-version = "0.1.46"
+version = "0.1.47"
 description = "HTTP client library for programmatic access to UiPath Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/packages/uipath-platform/src/uipath/platform/connections/_connections_service.py
+++ b/packages/uipath-platform/src/uipath/platform/connections/_connections_service.py
@@ -797,12 +797,22 @@ class ConnectionsService(BaseService):
                 # instead of making assumptions on whether or not it's present, we'll handle it defensively
                 if key == json_section:
                     continue
-                # files not supported yet supported so this will likely not work
-                files[key] = (
-                    key,
-                    val,
-                    None,
-                )  # probably needs to extract content type from val since IS metadata doesn't provide it
+                if isinstance(val, tuple):
+                    # Caller supplied httpx's (filename, content[, content_type])
+                    # shape — pass through verbatim. This is the recommended path
+                    # for file uploads so the multipart Content-Disposition gets
+                    # the real filename instead of the form-field name.
+                    files[key] = val
+                elif isinstance(val, (bytes, bytearray)) or hasattr(val, "read"):
+                    # Raw file content with no filename — fall back to the
+                    # form-field name (legacy behaviour). Backwards compatible
+                    # with callers that still pass bytes directly.
+                    files[key] = (key, val, "application/octet-stream")
+                else:
+                    # Scalar (string/number/etc.) — send as a plain multipart
+                    # form field, not a file part. The (None, value) shape tells
+                    # httpx to omit `filename=...` from the Content-Disposition.
+                    files[key] = (None, str(val))
 
             files[json_section] = (
                 "",
--- a/packages/uipath-platform/tests/services/test_connections_service.py
+++ b/packages/uipath-platform/tests/services/test_connections_service.py
@@ -2173,3 +2173,173 @@ class TestConnectorActivityInvocation:
         assert f"/element/instances/{original_connection_id}/" not in str(
             activity_request.url
         )
+
+
+def _multipart_part(body: bytes, boundary: str, name: str) -> str:
+    """Return the raw text of the multipart part with the given form-field name."""
+    text = body.decode("utf-8", errors="replace")
+    for part in text.split(f"--{boundary}"):
+        if f'name="{name}"' in part:
+            return part
+    raise AssertionError(f"part {name!r} not found in multipart body")
+
+
+class TestMultipartFileUpload:
+    """Regression tests for the multipart serializer that handles file uploads.
+
+    Before this fix, ``_build_activity_request_spec`` always built
+    ``files[key] = (key, val, None)``, using the form-field name as the
+    multipart filename and dropping the content type. Downstream services
+    (e.g. Coupa's ``add_attachment`` endpoint) ended up storing every
+    attachment with the literal name ``attachment[file]`` and no extension.
+
+    The serializer now branches on the value type:
+
+    * tuple  → passed through (caller controls filename + content type)
+    * bytes  → legacy fallback, key as filename, octet-stream content type
+    * scalar → plain multipart form field (no filename in Content-Disposition)
+    """
+
+    def test_invoke_activity_multipart_tuple_3_preserves_filename(
+        self,
+        httpx_mock: HTTPXMock,
+        service: ConnectionsService,
+        multipart_activity_metadata: ActivityMetadata,
+    ) -> None:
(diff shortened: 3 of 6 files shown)


radu-mocanu (approved):

Dennis-UiPath:
Fair pushback — three branches is the simplification that fits today's callers, not an exhaustive model of `multipart_params` value types.

**What's actually possible in `activity_input[key]` for a multipart key**

`multipart_params: Dict[str, Any]` — anything. The branches in this PR cover what current connector activities pass:

- **bytes / file-like** — file content (e.g. `attachment[file]: file_bytes`).
- **str (or other scalar)** — JSON-stringified metadata fields (e.g. `payload: \"{}\"`).
- **tuple** — new opt-in shape so callers can supply the real filename + content type.

**What httpx's `files=` actually accepts** (its `FileTypes`):
- `bytes`, `str`, `IO[AnyStr]` — file content
- `(filename, content)`, `(filename, content, content_type)`, or `(filename, content, content_type, headers)` — 2/3/4-tuples
- `Sequence[Tuple[str, FileTypes]]` for repeated form fields (multi-file uploads)

So things this PR handles **silently as scalars** that httpx would natively support if dispatched correctly:

- `memoryview` — coerced via `str()` to `\"<memory at 0x…>\"`. Trivial to fix by adding it to the bytes branch.
- 4-tuple with headers — works (passed through verbatim by the tuple branch).
- `list`/sequence for multi-value form fields — falls into the scalar branch and gets `str()`-ified. Not currently used by any IS connector activity I can find, but a real gap.
- `pathlib.Path` — coerced to its string path. The intent is almost certainly \"upload this file\", but we'd silently send the path as a form field.

**The scalar fallback is consistent with the rest of the function**: `query_params` / `path_params` / `header_params` already do `value_str = str(value) if not isinstance(value, str) else value` (line ~750). Multipart was the outlier that didn't coerce, and this PR brings it in line — which is also why I didn't want to make the unknown-type case raise (would change behaviour for any current caller passing an int/bool).

**Options, in increasing strictness:**

1. **Leave as-is.** Documented-but-narrow contract: tuple/bytes/file-like/scalar. Risky shapes silently mis-coerce.
2. **Widen the file branch** to also accept `bytearray|memoryview`. One-line change, no breaking impact, kills the most likely silent-corruption case.
3. **Add list / sequence handling** so multi-value uploads route correctly to httpx. Slightly more code; useful only if a connector ever needs it.
4. **Raise on unsupported types** instead of falling back to `str()`. Cleanest contract but changes behaviour for any caller passing a non-string scalar today.

I'd lean **(2)** as a follow-up tweak in this PR (one-liner, almost free), and leave (3) for whenever a multi-value multipart connector actually shows up. Happy to take (4) if the team prefers a strict contract — let me know.

Dennis-UiPath:
After rebasing the PR on main two tests fail. 

 Root cause (same in both jobs): the multimodal-invoke testcase tries to instantiate a model named gemini-3-pro-preview, and client_settings.get_model_info() raises:

```
  ValueError: Model gemini-3-pro-preview not found.
  Available models are: [..., 'gemini-2.5-flash', 'gemini-2.5-pro', ...]
```
  The available list on alpha contains gemini-2.5-flash and gemini-2.5-pro — but not gemini-3-pro-preview. Staging still has the model registered (those jobs pass), and the same alpha jobs passed for this PR and  PR #1614
  a few days ago. So between then and now somebody de-registered gemini-3-pro-preview from the alpha LLM Gateway catalog (or it was a time-limited preview registration that expired).

PR #1612, opened 2026-05-04
feat(platform): add W3C trace context headers to LLM Gateway requests

ctiliescuuipath:
## Summary
- Add `x-uipath-traceparent-id` and `x-uipath-tracebaggage` headers to all LLM Gateway HTTP calls
- Headers follow the [W3C Trace Context](https://www.w3.org/TR/trace-context/) spec, per [TraceView client SDK docs](https://uipath.atlassian.net/wiki/spaces/IDP/pages/90314834252)
- Gated behind the `EnableTraceContextHeaders` feature flag (env var: `UIPATH_FEATURE_EnableTraceContextHeaders=true`)
- Baggage includes: `source=agents`, `folderKey`, `agentId`, `processKey`
- `build_trace_context_headers()` defined once in `llm_trace_context.py`, consumed by both `UiPathLlmChatService`/`UiPathOpenAIService` and `uipath-langchain` legacy clients
- Bump `uipath-platform` version to `0.1.41`

## Test plan
- [x] Unit tests added in `test_llm_trace_context.py` (9 tests covering flag on/off, traceparent format, baggage content)
- [x] Existing LLM service tests pass (no regressions)
- [ ] Manual test with playground agent using `UIPATH_FEATURE_EnableTraceContextHeaders=true`

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
+++ b/packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
@@ -1,19 +1,19 @@
 """W3C-style trace context headers for LLM Gateway requests."""
 
-import os
-
 from opentelemetry import trace
 from uipath.core.feature_flags import FeatureFlags
 
-from ..common.constants import (
-    ENV_FOLDER_KEY,
-    ENV_PROCESS_KEY,
-    ENV_UIPATH_PROCESS_UUID,
-)
+from ..common._config import UiPathConfig
 
 
-def build_trace_context_headers() -> dict[str, str]:
+def build_trace_context_headers(
+    extra_baggage: list[str] | None = None,
+) -> dict[str, str]:
     """Build W3C-style trace context headers from the current OpenTelemetry span.
+
+    Args:
+        extra_baggage: Additional baggage entries (e.g. ``["source=agents"]``)
+            that callers can inject alongside the platform-level entries.
 
     Returns an empty dict when the ``EnableTraceContextHeaders`` feature flag
     is not enabled, or when no active span is present.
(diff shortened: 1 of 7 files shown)


packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
```diff
@@ -0,0 +1,41 @@
...
+
+    headers: dict[str, str] = {}
+    span = trace.get_current_span()
+    ctx = span.get_span_context()
+    if ctx and ctx.trace_id and ctx.span_id:
+        trace_id = format(ctx.trace_id, "032x")
+        span_id = format(ctx.span_id, "016x")
+        headers["x-uipath-traceparent-id"] = f"00-{trace_id}-{span_id}"
+
+    baggage_parts: list[str] = ["source=agents"]
```
radu-mocanu: this part should be moved to `uipath-langchain` and the other integration repos.
`uipath-platform` should remain framework/runtime agnostic (since function projects will also use it)  

```suggestion
    baggage_parts: list[str]
```

packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
```diff
@@ -0,0 +1,41 @@
...
+    if folder_key := os.getenv(ENV_FOLDER_KEY):
+        baggage_parts.append(f"folderKey={folder_key}")
+    if agent_id := os.getenv(ENV_UIPATH_PROCESS_UUID):
```
radu-mocanu:
```suggestion
    if agent_id := UiPathConfig.process_uuid:
```
let s use `UiPathConfig` here instead of using direct env vars accessors. we can even cache the data at that layer.

please note that some properties may not be available at the moment in `UiPathConfig`, we should add those

radu-mocanu (approved):

packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
```diff
@@ -0,0 +1,42 @@
...
+            that callers can inject alongside the platform-level entries.
+
+    Returns an empty dict when the ``EnableTraceContextHeaders`` feature flag
+    is not enabled, or when no active span is present.
+    """
+    if not FeatureFlags.is_flag_enabled("EnableTraceContextHeaders"):
+        return {}
+
+    headers: dict[str, str] = {}
+    span = trace.get_current_span()
```
ionmincu, line 25: you should use:
```
    from uipath.core.tracing.span_utils import UiPathSpanUtils

    llmops_span = UiPathSpanUtils.get_external_current_span()
    span = llmops_span or trace.get_current_span()
```

similar to https://github.com/UiPath/uipath-python/pull/1547/changes

PR #1614, opened 2026-05-05
fix: pass job key to is activities for licensing

andreitava-uip:
GenAI activities use `x-uipath-job-id` (which should contain the Job Key) for licensing reasons.

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/connections/_connections_service.py
+++ b/packages/uipath-platform/src/uipath/platform/connections/_connections_service.py
@@ -8,7 +8,7 @@ from uipath.core.tracing import traced
 
 from ..common._base_service import BaseService
 from ..common._bindings import resource_override
-from ..common._config import UiPathApiConfig
+from ..common._config import UiPathApiConfig, UiPathConfig
 from ..common._execution_context import UiPathExecutionContext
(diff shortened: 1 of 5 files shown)


packages/uipath-platform/src/uipath/platform/connections/_connections_service.py
```diff
@@ -768,11 +777,13 @@ def _build_activity_request_spec(
...
         # header parameter handling
         headers = {
-            "x-uipath-originator": "uipath-python",
-            "x-uipath-source": "uipath-python",
+            HEADER_ORIGINATOR: _ORIGINATOR_VALUE,
+            HEADER_SOURCE: _ORIGINATOR_VALUE,
             **header_folder(folder_key, None),
             **header_params,
         }
+        if job_key := os.getenv(ENV_JOB_KEY):
```
dianagrecu-uipath: nit: there is a method already defined in _config.py that wraps the os.getenv() call: https://github.com/UiPath/uipath-python/blob/main/packages/uipath-platform/src/uipath/platform/common/_config.py#L153
maybe we should use it instead

dianagrecu-uipath (approved):

PR #1616, opened 2026-05-07
feat: expand Data Fabric entities service [DS-8360]

avichalsri24:
## Summary

Adds the missing Data Fabric methods on `EntitiesService` and exposes the full backend parameter set on existing batch methods. Internally splits the service for clearer separation between schema- and data-side operations while keeping the public `sdk.entities.*` surface unchanged.

Jira: https://uipath.atlassian.net/browse/DS-8360

### Internal structure

`EntitiesService` is now a thin facade that delegates to two internal sub-services:

| Sub-service (internal) | Backend target | Responsibilities |
|---|---|---|
| `EntitySchemaService` (in `_entity_schema_service.py`, not exported) | `datafabric_/api/Entity` | entity / choiceset retrieval + create / delete / metadata update |
| `EntityDataService` (in `_entity_data_service.py`, not exported) | `datafabric_/api/EntityService/...` + `Attachment` + `v1/query/execute` | record CRUD (single + batch), structured query, FQS SQL query, attachments, bulk import, choiceset values |

Sub-services are deliberately not exported via `__init__.py` — `sdk.entities.*` remains the only sanctioned entry point. `EntityDataService.query` is the new structured-query method (named `query` so it can't be confused with the existing FQS `query_entity_records`, which is fully intact).

### New methods (sync + async on the facade)

- Single-record ops: `insert_record`, `get_record`, `update_record`, `delete_record` — fire trigger events on each mutation.
- `query` — structured query with `filter_group`, `sort_options`, `selected_fields`, `expansions`, `expansion_level`, `aggregates`, `group_by`, `joins`, `binnings`, `start`, `limit`. Routes to V2 endpoint when `binnings` is supplied.
- Attachments: `upload_attachment` (bytes or path), `download_attachment`, `delete_attachment`.
- Schema: `create_entity` (with full SQL-type mapping and per-type constraint defaults), `delete_entity`, `update_entity_metadata`.
- `import_records` for CSV bulk upload.

### Existing methods extended

- `insert_records` / `update_records` / `delete_records` accept `expansion_level` and `fail_on_first`.
- `list_records` accepts OData `filter` / `orderby` / `select` / `expand` / `expansion_level` and returns `EntityRecordsListResponse` (a `list` subclass with `total_count` / `has_next_page` / `next_cursor`).

### Bug fixes

- Batch operations recover per-record failures from HTTP 400 responses that carry `successRecords` / `failureRecords` lists; other non-2xx statuses propagate.
- Record input is normalized — accepts dicts, Pydantic models, `EntityRecord`, or any object with `__dict__`.

### Client-side validation

`create_entity` validates entity / field names client-side using the UI's create-form rules: `^[a-zA-Z][a-zA-Z0-9]*$` (no underscores), entity 1–30 chars, field 3–100 chars, plus reserved-name and per-field constraint range checks. Anything accepted by the SDK round-trips cleanly through the Data Service UI.

### Backward compatibility

Public method signatures only gained optional kwargs. `EntityRecord.id` stays required. `list_records` return type subclasses `list`, so iteration / indexing / `len()` / `isinstance(records, list)` continue to work. Existing docstrings are preserved verbatim from `main`; only ticket-mandated additions (new param entries, new Examples, new return types) appear in their docs.

## Test plan

- [x] `pytest packages/uipath-platform/tests/services/test_entities_service.py` — 112 pass
- [x] `pytest packages/uipath-platform/tests/` — 1158 pass
- [x] `pytest packages/uipath/tests/` — 1840 pass
- [x] `ruff check`, `ruff format --check`, `mypy src tests`, custom `lint_httpx_client.py` — clean across both packages
- [ ] Live smoke against a Data Fabric tenant before release tag


packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -25,15 +30,39 @@
...
     EntitySetResolution,
     QueryRoutingOverrideContext,
 )
 
+DATA_FABRIC_TENANT_FOLDER_ID = "00000000-0000-0000-0000-000000000000"
+
+FileContent = Union[bytes, bytearray, memoryview]
+"""Acceptable raw bytes types for attachment/CSV uploads."""
+
+_NAME_RE = re.compile(r"^[a-zA-Z]\w*$")
```
UIPath-Harshit: Is this consistent with backend?

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -864,17 +944,32 @@ def __init__(self, name, email, age):
...
-        response = await self.request_async(spec.method, spec.endpoint, json=spec.json)
+        spec = self._insert_batch_spec(
+            entity_key,
+            records,
+            expansion_level=expansion_level,
+            fail_on_first=fail_on_first,
+        )
 
-        return self.validate_entity_batch(response, schema)
+        async def _do() -> Response:
```
UIPath-Harshit: Better name here.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -1242,35 +1853,674 @@ def _query_entity_records_spec(
...
+    @staticmethod
+    def _update_entity_metadata_spec(
+        entity_id: str,
+        metadata: Union[EntityMetadataUpdateOptions, Dict[str, Any]],
+    ) -> RequestSpec:
+        if isinstance(metadata, EntityMetadataUpdateOptions):
+            body = metadata.model_dump(by_alias=True, exclude_none=True)
+        else:
+            body = dict(metadata)
+        return RequestSpec(
```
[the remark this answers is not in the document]
avichalsri24: dict inputs now route through EntityMetadataUpdateOptions.model_validate(...) then model_dump(by_alias=True, exclude_none=True), so display_name/is_rbac_enabled get serialized as displayName/isRbacEnabled on the wire. New test test_update_entity_metadata_normalizes_snake_case_dict_keys locks this in.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -1242,35 +1853,674 @@ def _query_entity_records_spec(
...
+            self._validate_name(field.field_name, "field")
+        opts = options or EntityCreateOptions()
+        payload: Dict[str, Any] = {
+            "displayName": opts.display_name or name,
+            "entityDefinition": {
+                "name": name,
+                "fields": [self._build_schema_field_payload(f) for f in fields],
+                "folderId": opts.folder_key or DATA_FABRIC_TENANT_FOLDER_ID,
+                "isRbacEnabled": bool(opts.is_rbac_enabled or False),
+                "isInsightsEnabled": bool(opts.is_analytics_enabled or False),
```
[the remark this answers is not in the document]
avichalsri24: is_analytics_enabled → isInsightsEnabled — added a one-line comment at the payload site explaining the legacy wire-name divergence.

packages/uipath-platform/src/uipath/platform/entities/entities.py
```diff
@@ -364,8 +384,352 @@ class EntityRecordsBatchResponse(BaseModel):
...
+        """Iterate over records (delegates to ``self.items``)."""
+        return iter(self.items)
+
+    def __len__(self) -> int:
+        """Return the number of records (delegates to ``self.items``)."""
+        return len(self.items)
+
+    def __getitem__(self, index: int) -> EntityRecord:
+        """Index records by position (delegates to ``self.items``)."""
+        return self.items[index]
```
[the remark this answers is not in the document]
avichalsri24: added @overload for int and slice on EntityQueryRecordsResponse, so slicing returns List[EntityRecord] and indexing returns EntityRecord with proper type narrowing.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -486,7 +545,12 @@ async def list_records_async(
...
-    ) -> List[EntityRecord]:
+        expansion_level: Optional[int] = None,
+        filter: Optional[str] = None,
+        orderby: Optional[str] = None,
+        select: Optional[List[str]] = None,
+        expand: Optional[List[str]] = None,
+    ) -> EntityRecordsListResponse:
         """Asynchronously list records from an entity with optional pagination and schema validation.
 
         The schema parameter enables type-safe access to entity records by validating the
```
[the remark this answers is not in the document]
avichalsri24: docstring now mirrors the sync version: Args list includes expansion_level, filter, orderby, select, expand; Returns describes EntityRecordsListResponse with pagination metadata; Examples include an OData filter/orderby/select/expand sample.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -1242,35 +1870,683 @@ def _query_entity_records_spec(
...
+    @staticmethod
+    def _delete_record_spec(entity_key: str, record_id: str) -> RequestSpec:
+        return RequestSpec(
+            method="DELETE",
+            endpoint=Endpoint(
+                f"datafabric_/api/EntityService/entity/{entity_key}/delete/{record_id}"
+            ),
+        )
+
+    def _query_records_spec(
```
UIPath-Harshit: !!! You are swapping out the query_records_spec which is catering to FQS query endpoint from DataService. This breaks multiple APIs for agents. Right strategy to avoid this is: 

1. Isolate the changes for schema into EntitySchemaService
2. Isolate the changes for data into EntityDataService. 
3. EntityService is a facade that doesnt change the external contract. 
4. Carefully understand the APIs for DS and FQS and then make these changes.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
avichalsri24: Updated to _send_batch

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
avichalsri24: Verified against the UI's create-form validators. Tightened _NAME_RE to ^[a-zA-Z][a-zA-Z0-9]*$ (no underscores) and split lengths by context: entity 1–30, field 3–100. Anything accepted by the SDK can now round-trip through the UI.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
avichalsri24: As discussed, this is a new method for odata query and not impacting sql endpoint consumed by agents (query_entity_records_spec vs query_records_spec). 


Regarding refactor, updated the code, split into EntitySchemaService (entity/choiceset CRUD) and EntityDataService (records, attachments, queries — both FQS query_entity_records and DS query). EntitiesService is now a thin facade that preserves the existing flat sdk.entities.* contract via explicit delegation. No external API change. The new DS structured-query method is named query (not query_records) so it can't be confused with the existing FQS query_entity_records, which is fully intact. Sub-services are kept internal (not exported in __init__.py).

avichalsri24:
@codex review

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+]
+_AGGREGATE_FUNCTIONS = ("COUNT", "SUM", "AVG", "MIN", "MAX")
+
+
+class EntityDataService(BaseService):
+    """HTTP service for entity-record and attachment operations.
+
+    Backend target: ``datafabric_/api/EntityService/...`` plus
+    ``datafabric_/api/Attachment/...`` for file attachments, and
+    ``datafabric_/api/v1/query/execute`` for legacy SQL queries.
```
UIPath-Harshit: not legacy :) rather thats the new api that will be used in the future.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+        result = await self._request_or_extract_batch_async(_send_batch)
+        if isinstance(result, EntityRecordsBatchResponse):
+            return result
+        return EntityRecordsBatchResponse.model_validate(result.json())
+
+    # ------------------------------------------------------------------
+    # Structured query (POST /entity/{id}/query)
+    # ------------------------------------------------------------------
+
+    def query(
```
UIPath-Harshit: lets mark this as retrieve_records to ensure that query endpoint is tied back to sql

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+        selected_fields: Optional[List[str]] = None,
+        expansions: Optional[List[Any]] = None,
+        expansion_level: Optional[int] = None,
+        aggregates: Optional[List[EntityAggregate]] = None,
+        group_by: Optional[List[str]] = None,
+        joins: Optional[List[EntityJoin]] = None,
+        binnings: Optional[List[EntityBinning]] = None,
+        start: Optional[int] = None,
+        limit: Optional[int] = None,
+    ) -> EntityQueryRecordsResponse:
```
UIPath-Harshit: RetrieveEntityRecordsResponse.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+            start=start,
+            limit=limit,
+        )
+        response = await self.request_async(
+            spec.method, spec.endpoint, params=spec.params, json=spec.json
+        )
+        return self._parse_query_response(response, start=start, limit=limit)
+
+    # ------------------------------------------------------------------
+    # Federated SQL query (legacy escape hatch)
```
UIPath-Harshit: Why are we calling this legacy ?

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+        self,
+        entity_id: str,
+        file: Optional[FileContent] = None,
+        file_path: Optional[str] = None,
+    ) -> EntityImportRecordsResponse:
+        """Internal implementation; see :meth:`EntitiesService.import_records`."""
+        with self._open_file(file, file_path) as handle:
+            response = self.request(
+                "POST",
+                Endpoint(
```
UIPath-Harshit: no spec method?

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+    ) -> List[Dict[str, Any]]:
+        """Asynchronously run a validated SQL query through the federated query engine."""
+        self._validate_sql_query(sql_query)
+        routing_context = await self._routing_strategy.resolve_async()
+        spec = self._query_entity_records_spec(sql_query, routing_context)
+        response = await self.request_async(spec.method, spec.endpoint, json=spec.json)
+        return response.json().get("results", [])
+
+    @staticmethod
+    def _list_records_spec(
```
UIPath-Harshit, line 702: Can we make the params pydantic using builder method?

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+        """Build the optional URL params common to all batch endpoints."""
+        params: Dict[str, Any] = {}
+        if expansion_level is not None:
+            params["expansionLevel"] = expansion_level
+        if fail_on_first is not None:
+            params["failOnFirst"] = "true" if fail_on_first else "false"
+        return params
+
+    @staticmethod
+    def _query_spec(
```
UIPath-Harshit: Same here. Can we pass pydantic models here also

UIPath-Harshit (commented):
Initial comments for data aspects.

packages/uipath-platform/src/uipath/platform/entities/entities.py
```diff
@@ -364,8 +385,358 @@ class EntityRecordsBatchResponse(BaseModel):
...
+    )
+    display_name: Optional[str] = Field(default=None, alias="displayName")
+    description: Optional[str] = None
+    is_required: Optional[bool] = Field(default=None, alias="isRequired")
+    is_unique: Optional[bool] = Field(default=None, alias="isUnique")
+    is_rbac_enabled: Optional[bool] = Field(default=None, alias="isRbacEnabled")
+    is_encrypted: Optional[bool] = Field(default=None, alias="isEncrypted")
+    default_value: Optional[str] = Field(default=None, alias="defaultValue")
+    length_limit: Optional[int] = Field(default=None, alias="lengthLimit")
+    max_value: Optional[float] = Field(default=None, alias="maxValue")
```
UIPath-Harshit: `EntityCreateFieldOptions.max_value` / `min_value` are typed as `float` for every field kind, and `_validate_field_constraints()` only checks range. That means `EntityCreateFieldOptions(type=INTEGER, max_value=3.5)` passes client-side validation and gets serialized as `3.5` for an integer column. Integer-backed field types should reject fractional bounds locally, or use a narrower type for those cases.

UIPath-Harshit: `create_entity()` now allows DECIMAL/FLOAT/DOUBLE fields to send fractional `maxValue` / `minValue`, but the retrieve-side `FieldDataType.max_value` / `min_value` model is still `Optional[int]`. If the backend returns something like `3.14` for one of those constraints, `retrieve()` / `list_entities()` will reject or coerce it incorrectly. The response model needs to accept non-integer numeric bounds if the request model can create them.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
```diff
@@ -0,0 +1,1401 @@
...
+        # Aggregate / binning rows do not carry an ``Id`` field, so fall back
+        # to constructing the record without strict validation when the row
+        # cannot be parsed as a regular entity record.
+        items_raw = body.get("value", []) or []
+        items: List[EntityRecord] = []
+        for raw in items_raw:
+            try:
+                items.append(EntityRecord.from_data(data=raw))
+            except ValueError:
+                items.append(EntityRecord.model_construct(_fields_set=set(raw), **raw))
```
UIPath-Harshit: This makes aggregate/binning rows look like `EntityRecord`s even though they may not have an `Id`. The parser gets the request through, but the declared return type is now lying: `result.items` is typed as `List[EntityRecord]`, while `row.id` can still raise `AttributeError` on these constructed rows. It would be safer to widen the item type here or introduce a separate response model for aggregate/binning queries instead of fabricating invalid `EntityRecord` instances.

packages/uipath-platform/src/uipath/platform/entities/entities.py
```diff
@@ -364,8 +385,358 @@ class EntityRecordsBatchResponse(BaseModel):
...
+    min_value: Optional[float] = Field(default=None, alias="minValue")
+    decimal_precision: Optional[int] = Field(default=None, alias="decimalPrecision")
+    choice_set_id: Optional[str] = Field(default=None, alias="choiceSetId")
```
UIPath-Harshit, line 760: Type-dependent required fields are not enforced here. `create_entity()` currently accepts a `RELATIONSHIP` field without `referenceEntityName` / `referenceFieldName`, and `CHOICE_SET_SINGLE` / `CHOICE_SET_MULTIPLE` fields without `choiceSetId`, then serializes those incomplete payloads to the backend. This needs validation in the new schema surface so callers get a local error instead of sending an invalid field definition.

packages/uipath-platform/src/uipath/platform/entities/entities.py
```diff
@@ -364,8 +385,358 @@ class EntityRecordsBatchResponse(BaseModel):
...
+
+
+class EntityQueryFilter(BaseModel):
+    """A single filter condition for querying entity records."""
+
+    model_config = ConfigDict(validate_by_name=True, validate_by_alias=True)
+
+    field_name: str = Field(alias="fieldName")
+    operator: QueryFilterOperator
+    value: Optional[str] = None
```
UIPath-Harshit, line 467: `EntityQueryFilter` currently accepts any combination of `value`, `valueList`, or neither, regardless of operator. That means malformed shapes like `In`/`NotIn` with only `value`, `Equals` with only `valueList`, or filters with no operand at all all validate locally and are serialized. The structured-query model should enforce scalar-vs-list operands by operator so invalid queries are rejected before the network call.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: Existing DS spec helpers (and every other platform service) use plain kwargs. Doing it just here would create two patterns in the same file. If we want to move to Pydantic options builders, it's worth doing consistently across the entities service in a follow-up and not as part of this change.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: Existing DS spec helpers (and every other platform service) use plain kwargs. Doing it just here would create two patterns in the same file. If we want to move to Pydantic options builders, it's worth doing consistently across the entities service in a follow-up and not as part of this change.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: Got it — dropped the "legacy" framing across the file (docstrings + section comments).

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: Renamed to `retrieve_records` / `retrieve_records_async` on both the facade and the data sub-service. The FQS endpoint stays as `query_entity_records` so the two surfaces don't get confused.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: Renamed. Updated the model, the data-service and facade signatures, the `__init__.py` exports, and the tests.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: My bad — removed the "legacy" wording here too.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: Good catch — extracted `_import_records_spec`. Both `import_records` and `import_records_async` go through it now, consistent with the rest of the file.

packages/uipath-platform/src/uipath/platform/entities/entities.py
avichalsri24: Switched `EntityCreateFieldOptions.max_value` / `min_value` from `float` to `int`. Python's `int` rejects fractional values during Pydantic validation, so `max_value=3.5` on any field now errors out locally without us needing an explicit check. Bounds are whole numbers backend-side anyway — fractional precision is controlled separately via `decimal_precision`.

avichalsri24: Both sides are aligned to `int` now — request and response. Since the create side no longer accepts fractional bounds, there's nothing fractional for the response side to round-trip.

packages/uipath-platform/src/uipath/platform/entities/_entity_data_service.py
avichalsri24: Fair point. Introduced a separate `AggregateRow` model (`extra="allow"`) and widened `items` to `List[EntityRecord | AggregateRow]`. The parser now dispatches on the presence of `Id`: rows with `Id` go through `EntityRecord`, the rest through `AggregateRow`. No more fabricating `EntityRecord` instances without an `id`.

packages/uipath-platform/src/uipath/platform/entities/entities.py
avichalsri24: Added the type-dependent checks in `_validate_field_constraints`: `CHOICE_SET_SINGLE` / `CHOICE_SET_MULTIPLE` now require `choice_set_id`, and `RELATIONSHIP` requires `reference_entity_name`. Callers get a local `ValueError` instead of sending an invalid payload. Covered by three new tests.

packages/uipath-platform/src/uipath/platform/entities/entities.py
avichalsri24: Added a `model_validator` on `EntityQueryFilter` that mirrors the rules `SelectQueryBuilder` enforces server-side:

- `IN` / `NOT IN` → require a non-empty `value_list`, reject `value`.
- `=` / `!=` → can have a null `value` (becomes `IS NULL` / `IS NOT NULL`), reject `value_list`.
- All other operators (`>`, `<`, `>=`, `<=`, `contains`, `not contains`, `startswith`, `endswith`) → require a non-null `value`, reject `value_list`.

Bad shapes now fail at construction time with a clear message; 6 new tests cover each branch.

UIPath-Harshit (approved):

PR #1624, opened 2026-05-11
feat(eval): add runtime simulation via --simulation flag

AAgnihotry:
## Summary

- Adds `--simulation` flag to `uipath run` that accepts a JSON config (same schema as `simulation.json`) to wrap the runtime with `UiPathMockRuntime`
- Exports `build_mocking_context_from_dict` from `uipath.eval.mocks` for use by the CLI
- Adds `runtime-simulations-agent` sample demonstrating `@mockable`-decorated tools with a `simulation.json`
- Adds `simulation-testcase` integration testcase with `run.sh` + `src/assert.py` that verifies simulation was triggered (log lines + non-default LLM output)
- Bumps version to `2.10.63`

## Test plan

- [x] All 1840 existing unit tests pass (`uv run pytest`)
- [x] No lint errors (`ruff check`, `ruff format --check`, `lint_httpx_client.py`)
- [x] Run sample locally: `cd packages/uipath/samples/runtime-simulations-agent && uv run uipath run main -f input.json --simulation "$(cat simulation.json)"`
- [x] Integration testcase runs via `integration_tests.yml` when `uipath` package changes

## Context notes
- Vertical Solutions team doesn't use Orchestrator.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


packages/uipath/src/uipath/eval/mocks/_mock_runtime.py
```diff
@@ -35,6 +35,53 @@
 logger = logging.getLogger(__name__)
 
 
+def build_mocking_context_from_dict(
```
Chibionos: **P1 — F1: Loose `dict[str, Any]` input; no Pydantic validation**

Pillar 15 — Python Service Discipline. The function accepts a raw dict and silently no-ops on missing/malformed keys. The repo already uses Pydantic v2 for the related shapes — `MockingContext`, `LLMMockingStrategy`, `ToolSimulation`, `ModelSettings` all live in `_types.py` with `validate_by_name=True, validate_by_alias=True`.

The `_from_dict` suffix is itself a smell: the function is shaped around the on-disk JSON schema. Make the schema a Pydantic model and pass *that* through:

```python
# In _types.py
class SimulationConfig(BaseModel):
    enabled: bool = True
    tools_to_simulate: list[ToolSimulation] = Field(default_factory=list, alias="toolsToSimulate")
    instructions: str = ""
    model: str | None = None

    model_config = ConfigDict(
        validate_by_name=True, validate_by_alias=True, extra="allow"
    )

# In _mock_runtime.py — rename `_from_dict` → `_from_config` and require typed input
def build_mocking_context(
    config: SimulationConfig, agent_model: str | None = None
) -> MockingContext | None:
    if not config.enabled or not config.tools_to_simulate:
        return None

    model = (
        ModelSettings(model=config.model) if config.model
        else ModelSettings(model=agent_model) if agent_model
        else None
    )
    strategy = LLMMockingStrategy(
        type=MockingStrategyType.LLM,
        prompt=config.instructions,
        tools_to_simulate=config.tools_to_simulate,
        model=model,
    )
    return MockingContext(strategy=strategy, name="debug-simulation", inputs={})


# In cli_run.py — validate at the entry, with a useful error message
if simulation:
    try:
        config = SimulationConfig.model_validate_json(simulation)
    except (json.JSONDecodeError, ValidationError) as e:
        console.error(f"Invalid --simulation config: {e}")
        return
    ...
    mocking_context = build_mocking_context(config, agent_model)
```

Why this matters: today, a typo like `--simulation '{"toolToSimulate": [...]}'` (singular) silently returns `None` and the agent runs without mocking. With Pydantic the user gets `ValidationError: missing field 'toolsToSimulate'` at parse time.

Chibionos (commented):
## /deep-review — overall

**Verdict: Comment.** Two P1s (one Pydantic validation gap, one runtime-lifecycle leak that Codex caught and I verified), two P2s, one P3. All findings are inline below — each can be resolved / replied to individually.

### Slop meter
🟢 **clean** — slop 4/100 · tunnel-vision 0/100. The 4 points are version literals in `pyproject.toml` files, expected for sample/testcase scaffolding.

### Codex recheck
**Verdict: amend** — Codex agreed with my four draft findings and surfaced one P1 I'd missed (runtime-lifecycle leak), now F2.

### Findings index (all inline)

|---|---|---|---|
| F1 | P1 | `_mock_runtime.py:38` | Loose `dict[str, Any]` input — no Pydantic validation |
| F2 | P1 | `cli_run.py:249` | Simulated `run` leaks the delegate runtime (Codex catch) |
| F3 | P2 | `cli_run.py:107` | `--simulation` accepts inline JSON only — awkward UX |
| F4 | P2 | `testcases/simulation-testcase/main.py` | Sample + testcase duplicate the same fixture |
| F5 | P3 | `_mock_runtime.py:77` | `logger.info` likely too chatty — consider `logger.debug` |

### Pillar 15 — Python Service Discipline

The deep-review skill grew a Python pillar today, distilled from 104 review comments by [@cristipufu](https://github.com/cristipufu) and [@radu-mocanu](https://github.com/radu-mocanu) on this repo over the last 120 days. F1 and F4 trace directly back to patterns those two flag repeatedly. The full pillar lives at `skills-internal#299` (`plugins/deep-review/skills/deep-review/references/default-pillars.md`).

packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -218,6 +235,22 @@ async def execute() -> None:
...
+                                agent_model = None
+                                if schema.metadata and "settings" in schema.metadata:
+                                    agent_model = schema.metadata["settings"].get(
+                                        "model"
+                                    )
+                                mocking_context = build_mocking_context_from_dict(
+                                    simulation_data, agent_model
+                                )
+                                if mocking_context:
+                                    runtime = UiPathMockRuntime(
```
Chibionos, line 251: **P1 — F2: Simulated `run` leaks the delegate runtime** _(caught by Codex recheck)_

This line reassigns `runtime` to a `UiPathMockRuntime` wrapper. The `finally` block at line 277 then calls `await runtime.dispose()` — but `UiPathMockRuntime.dispose()` is a deliberate no-op (`_mock_runtime.py:208`):

```python
async def dispose(self) -> None:
    """No-op; callers manage delegate disposal."""
    pass
```

Which means: on a simulated run, the original delegate runtime's `dispose()` is never called. Whatever the base runtime holds (DB connections, file handles, span processors, child processes) leaks until the process exits. Non-simulated runs are unaffected.

Fix — keep a reference to the base runtime and dispose both:

```python
base_runtime = await factory.new_runtime(
    resolved_entrypoint,
    ctx.conversation_id or ctx.job_id or "default",
)
runtime = base_runtime

if simulation_data:
    schema = await base_runtime.get_schema()
    agent_model = None
    if schema.metadata and "settings" in schema.metadata:
        agent_model = schema.metadata["settings"].get("model")

    mocking_context = build_mocking_context_from_dict(simulation_data, agent_model)
    if mocking_context:
        runtime = UiPathMockRuntime(
            delegate=base_runtime,
            mocking_context=mocking_context,
        )

# ... existing execute path uses `runtime`

# in finally:
if chat_runtime:
    await chat_runtime.dispose()
if runtime is not None and runtime is not base_runtime:
    await runtime.dispose()
if base_runtime is not None:
    await base_runtime.dispose()
```

Alternative: change `UiPathMockRuntime.dispose()` to call `await self.delegate.dispose()`. That's a smaller diff but couples wrapping policy with cleanup policy — riskier.

packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -101,6 +103,12 @@ def get_usage_help(self) -> list[str]:
     is_flag=True,
     help="Keep the temporary state file even when not resuming and no job id is provided",
 )
+@click.option(
+    "--simulation",
```
Chibionos, line 107: **P2 — F3: `--simulation` accepts inline JSON only — awkward UX**

Forcing users to embed raw JSON in a command line is brittle. Single quotes inside double-quoted JSON, Windows `cmd.exe` quoting differences, shell-injection footguns. Sibling option `--state-file` already shows the pattern with `click.Path(exists=True)`.

Three options, in increasing order of effort:

1. **Rename to `--simulation-config` and accept a file path only** — cleanest, follows the rest of the CLI:

```python
@click.option(
    "--simulation-config",
    required=False,
    type=click.Path(exists=True, dir_okay=False),
    help="Path to a simulation config JSON file (same schema as simulation.json)",
)
```

2. **Accept either inline JSON or `@path/to/file.json` (cURL convention)**:

```python
if simulation:
    if simulation.startswith("@"):
        with open(simulation[1:], "r") as f:
            simulation = f.read()
    try:
        config = SimulationConfig.model_validate_json(simulation)
    except (json.JSONDecodeError, ValidationError) as e:
        console.error(f"Invalid --simulation config: {e}")
        return
```

3. **Keep as-is**, only if the inline form is the dominant expected use case (CI scripts where JSON is generated and piped in). If that's the rationale, please add a one-line `help` hint about Windows quoting.

The sample in `samples/runtime-simulations-agent/main.py` already documents the inline form as `uipath run ... --simulation "$(cat simulation.json)"` — which itself is a workaround for the missing file-path mode.

packages/uipath/testcases/simulation-testcase/main.py
```diff
@@ -0,0 +1,77 @@
+from pydantic import BaseModel
```
Chibionos: **P2 — F4: Sample (`samples/runtime-simulations-agent/`) and testcase (`testcases/simulation-testcase/`) duplicate the same fixture**

Both ship a 3-tool code-review agent (`check_syntax`, `check_style`, `suggest_improvements`) with near-identical `simulation.json` files (verbatim except for two trimmed phrases in the `instructions` string). Same input shape, same mock tools, same `@mockable` decorations.

Two options:

1. **Drop the testcase's own `main.py` / `simulation.json` and have its `run.sh` execute against `samples/runtime-simulations-agent/`**:

```bash
# testcases/simulation-testcase/run.sh
#!/bin/bash
set -e
cd ../../samples/runtime-simulations-agent
uipath run main.py:main -f input.json --simulation "$(cat simulation.json)" | tee /tmp/sim-run.log
python3 -c "..." /tmp/sim-run.log  # assertions
```

2. **Keep both but document the why** — add a 3-line `README.md` to the testcase explaining what it tests that the sample doesn't (probably: the simulation actually triggered and produced non-default output). The duplication itself isn't fatal, but a future contributor will inevitably edit one and not the other; document the intent so they know which to sync.

This is a Pillar 2 (Reuse) finding cross-tagged with Pillar 15 (Python — testing discipline).

packages/uipath/src/uipath/eval/mocks/_mock_runtime.py
```diff
@@ -35,6 +35,53 @@
...
+    )
+
+    mocking_strategy = LLMMockingStrategy(
+        type=MockingStrategyType.LLM,
+        prompt=simulation_data.get("instructions", ""),
+        tools_to_simulate=tools_to_simulate,
+        model=model,
+    )
+
+    logger.info(f"Loaded simulation config for {len(tools_to_simulate)} tool(s)")
```
Chibionos: **P3 — F5: `logger.info` for simulation config load — likely too chatty**

This is on the hot path of every simulated run. In the corpus of @cristipufu's prior PR comments on this repo, this exact pattern came up: *"I think we should remove these logs, they were useful for debugging, but for the end-user not so useful"*. Borderline since the simulation is user-initiated, but the line count + tool list is already visible from the user's own `simulation.json`.

```diff
- logger.info(f"Loaded simulation config for {len(tools_to_simulate)} tool(s)")
+ logger.debug(f"Loaded simulation config for {len(tools_to_simulate)} tool(s)")
```

P3 — flagged once, don't open inline threads on every similar `logger.info` in the file. Take it or leave it.

packages/uipath/src/uipath/_cli/cli_run.py
AAgnihotry: This is by design, we dont want users to upload the file

packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -114,6 +122,7 @@ def run(
     debug: bool,
     debug_port: int,
     keep_state_file: bool,
+    simulation: str | None,
```
mjnovice, line 125: Why is there a need for specifying --simulation as a separate flag ? Isn't the current behaviour implicitly based on the presence of simulation.json file like we have for the debug runs https://github.com/UiPath/uipath-python/pull/1117 ?

AAgnihotry: we dont want to use simulations.json debug time at publish time. we should be able to configure it at runtime

Chibionos (approved):

PR #1626, opened 2026-05-12
fix(tracing): set Traces source to CodedAgents

ionmincu:
## Summary
- Change `source=Robots` to `source=CodedAgents` in the `/api/Traces/spans` URL built by `LlmOpsHttpExporter`
- Update related tests
- Bump `uipath` package version to 2.10.63

## Test plan
- [ ] `pytest packages/uipath/tests/tracing/test_otel_exporters.py`

### DIFF
--- a/packages/uipath/pyproject.toml
+++ b/packages/uipath/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.10.62"
+version = "2.10.63"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/packages/uipath/src/uipath/tracing/_otel_exporters.py
+++ b/packages/uipath/src/uipath/tracing/_otel_exporters.py
@@ -389,7 +389,7 @@ class LlmOpsHttpExporter(SpanExporter):
     def _build_url(self, span_list: list[Dict[str, Any]]) -> str:
         """Construct the URL for the API request."""
         trace_id = str(span_list[0]["TraceId"])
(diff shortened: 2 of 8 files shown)


packages/uipath/src/uipath/tracing/_otel_exporters.py
```diff
@@ -389,7 +389,7 @@ def _process_span_attributes(self, span_data: Dict[str, Any]) -> None:
     def _build_url(self, span_list: list[Dict[str, Any]]) -> str:
         """Construct the URL for the API request."""
         trace_id = str(span_list[0]["TraceId"])
-        return f"{self.base_url}/api/Traces/spans?traceId={trace_id}&source=Robots"
```
radu-mocanu, line 392: are we sure this won t break low code?
we need to double-check that all the spans we emit are dropped in by their runtime

cotovanu-cristian: @saksharthakkar / @JosephMar could validate this change from a tracing perspective

JosephMar: This also needs to be changed https://github.com/UiPath/uipath-python/blob/dc5319f25c52581a79b2f9512b8915627c634509/packages/uipath-platform/src/uipath/platform/common/_span_utils.py#L19
source exists both inside the model and on the query path (for convenience)

packages/uipath/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.10.62"
```
cotovanu-cristian, line 3: shouldn't we also re-generate the `uv.lock`?

ionmincu: probably

ionmincu:
reopened https://github.com/UiPath/uipath-python/pull/1631, cant seem to push to this branch

radugheo (approved):

PR #1632, opened 2026-05-15
feat(eval): add --simulation flag to uipath debug

ajay-kesavan:
## Summary

Mirrors the `--simulation` flag introduced for `uipath run` in #1624 onto `uipath debug`, so a simulation config can be passed inline (same JSON schema as `simulation.json`) without writing it to disk.

- Adds `--simulation` option + parameter to `cli_debug.py`
- Validates the JSON via `SimulationConfig.model_validate_json` (uses `pydantic.ValidationError` for surfacing errors)
- When `--simulation` is provided, builds the mocking context via `build_mocking_context(simulation_config, agent_model)`
- When `--simulation` is omitted, the existing `load_simulation_config()` fallback (reads `simulation.json` from cwd) is preserved — this is the one behavioral difference from `run`, which has no file-based fallback

## Test plan

- [x] `ruff check` and `ruff format --check` pass on the modified file
- [ ] Unit tests for `debug --simulation` (mirroring the `TestRunSimulation` class added to `tests/cli/test_run.py` in #1624) are **not** included in this PR — happy to add in a follow-up if reviewers want them here

## Scope notes

This PR only mirrors the CLI flag. The sample (`runtime-simulations-agent`) and integration testcase (`simulation-testcase`) added in #1624 exercise `uipath run` and are not duplicated for `debug` since the underlying `build_mocking_context` plumbing they cover is shared.

### DIFF
--- a/packages/uipath/src/uipath/_cli/cli_debug.py
+++ b/packages/uipath/src/uipath/_cli/cli_debug.py
@@ -3,13 +3,18 @@ import logging
 from typing import cast, get_args
 
 import click
+from pydantic import ValidationError
 
 from uipath._cli._chat._bridge import get_chat_bridge
 from uipath._cli._debug._bridge import DebugAttachMode, get_debug_bridge
 from uipath._cli._utils._debug import setup_debugging
 from uipath._cli._utils._studio_project import StudioClient
 from uipath.core.tracing import UiPathTraceManager
-from uipath.eval.mocks import UiPathMockRuntime
+from uipath.eval.mocks import (
+    SimulationConfig,
+    UiPathMockRuntime,
+    build_mocking_context,
+)
 from uipath.eval.mocks._mock_runtime import load_simulation_config
 from uipath.platform.common import ResourceOverwritesContext, UiPathConfig
 from uipath.runtime import (
@@ -74,6 +79,12 @@ logger = logging.getLogger(__name__)
         "'console' for local runs."
     ),
 )
+@click.option(
+    "--simulation",
+    required=False,
+    default=None,
+    help="Simulation config as a JSON object (same schema as simulation.json)",
+)
 @track_command("debug")
 def debug(
     entrypoint: str | None,
@@ -85,6 +96,7 @@ def debug(
     debug: bool,
     debug_port: int,
     attach: str | None,
+    simulation: str | None,
 ) -> None:
     """Debug the project."""
     input_file = file or input_file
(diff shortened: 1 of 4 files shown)


AAgnihotry:
can you please add some tests here

ajay-kesavan:
https://github.com/user-attachments/assets/7a005fe5-40c4-4e19-b110-e2712e243f08

ajay-kesavan:
[simulation-agent.zip](https://github.com/user-attachments/files/28033508/simulation-agent.zip)

packages/uipath/uv.lock
```diff
@@ -3,7 +3,7 @@ revision = 3
 requires-python = ">=3.11"
 
 [options]
-exclude-newer = "2026-05-17T16:21:47.0725551Z"
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
```
Chibionos, line 6: ### 🔵 `exclude-newer` flipped to epoch zero
`exclude-newer` changed from the real timestamp `2026-05-17T...` to `0001-01-01T00:00:00Z`. This is a uv-version artifact and is harmless, but confirm it came from an intentional `uv lock` regeneration rather than a stray downgrade of the uv toolchain — it can subtly change dependency resolution on the next lock.

packages/uipath/src/uipath/_cli/cli_debug.py
```diff
@@ -96,6 +108,14 @@ def debug(
         cast(DebugAttachMode, attach.lower()) if attach else None
     )
 
+    simulation_config: SimulationConfig | None = None
+    if simulation:
+        try:
+            simulation_config = SimulationConfig.model_validate_json(simulation)
+        except (ValidationError, ValueError) as e:
+            console.error(f"Invalid --simulation config: {e}")
+            return
```
Chibionos, line 117: ### 🔵 unreachable `return`
`console.error()` is `NoReturn` (raises `SystemExit`), so this `return` never executes. Harmless and consistent with the existing pattern at `:138`; fine to leave. Flagging only so it's a conscious choice.

Chibionos (commented):
## PR #1632 Review — `--simulation` flag on `uipath debug`

**Verdict: ✅ Approve after rebase.** Clean mirror of the `uipath run --simulation` flag (#1624). I checked out the head, ran the suite, and traced the flow.

### ✅ Verified working
- **18/18 tests pass** (`tests/cli/test_debug_simulation.py`), including the 3 new `TestDebugSimulationFlag` tests.
- `ruff check` + `ruff format --check` clean.
- Closure capture of `simulation_config` into the nested `execute_debug_runtime` is correct.
- `except (ValidationError, ValueError)` matches `cli_run.py` exactly; the `console.error()` early-exit (`NoReturn` → `SystemExit`) correctly prevents reaching `Middlewares.next` — and the test's `assert not factory_get.called` confirms it.
- All five `patch(...)` targets resolve to real names in `cli_debug`'s namespace.

### 🔵 Before merge
- Branch is `CONFLICTING` and bumps `2.10.67 → 2.10.68`; rebase and resolve the version-line conflict (note #1674 also bumps the version).

Two minor inline observations below — nothing blocking.

---
*Adversarial + deep-review pass; verified on the PR head.*

PR #1638, opened 2026-05-18
feat: add client side tools to bridge and new cas events [JAR-9629]

norman-le:
Had to make new PR (old PR here: https://github.com/UiPath/uipath-python/pull/1609)

```
Changes across repos for client-side tools feature:
uipath-python:
- Removed toolName from ExecutingToolCallEvent model. 
- Added client-side tool declaration model on exchange start. 
- Bridge gains emit_executing_tool_call method and wait_for_resume buffering for early responses.

uipath-runtime-python:
- Added emit_executing_tool_call to protocol. 
- Runtime loop emits executingToolCall after tool-call confirmations resume.

uipath-langchain-python:
- Client-side tool factory with @durable_interrupt. Names-only tool filtering via apply_tool_filter. 
- MessageMapper emits executingToolCall for all non-confirmed tools, adds isClientSideTool/outputSchema to startToolCall, suppresses endToolCall for client-side results. 
- Runtime discovers client-side tools through graph introspection. 
- Init node applies per-exchange filtering.

uipath-agents-python:
- Strips client-side tool metadata during evals.

uipath-dev-python:
- Protocol stub for emit_executing_tool_call.

AgentInterfaces:
- Schema updates (ExecutingToolCallEvent, ClientSideToolDeclaration, exchange start). 
- CAS passes client-side tool names to runtime input, forwards confirmToolCall to CLI bridge. 
- React SDK handles executingToolCall for client-side tools.
```

Videos in CAS localhost (Agent Builder changes already published, but this will be the debug experience as well when the sdk is upgraded):

Without tool confirmation for client side tool:

https://github.com/user-attachments/assets/48bc0979-0f51-471c-90aa-af561d876c31

With tool confirmation for client side tool:

https://github.com/user-attachments/assets/384ba36d-db2d-478e-8a74-8b3de0501a06

URT Eval:
image

Related to changes in other PRs:
CAS: https://github.com/UiPath/AgentInterfaces/pull/949
uipath-langchain-python (message mapping and firing executing tool call if no confirmation): https://github.com/UiPath/uipath-langchain-python/pull/856
uipath-python (definitions): https://github.com/UiPath/uipath-python/pull/1638
uipath-agents-python (skipping for evals, so it can be mocked): https://github.com/UiPath/uipath-agents-python/pull/485
uipath-runtime-python (protocol changing to emit executing tool call event after confirmation): https://github.com/UiPath/uipath-runtime-python/pull/115
Another Agents PR fixing some issues with debug: https://github.com/UiPath/Agents/pull/5250


packages/uipath/src/uipath/_cli/_chat/_bridge.py
```diff
@@ -362,33 +367,49 @@ async def emit_exchange_error_event(self, error: Exception) -> None:
...
+        value = self._tool_resume_value
+        self._tool_resume_value = None
+        self._tool_resume_event.clear()
 
-        if self._tool_confirmation_value:
-            return self._tool_confirmation_value.model_dump(
-                mode="python", by_alias=False
-            )
+        if value:
+            return value.model_dump(mode="python", by_alias=False)
```
maxduu, line 415: Curious why we have the check for `None`, and then a clear when its `None`. Shouldn't the overall logic be identical to before, only difference being some renames because it can be either a confirmation-event or end-event? I might be missing something though

packages/uipath/src/uipath/_cli/_chat/_bridge.py
```diff
@@ -362,33 +367,49 @@ async def emit_exchange_error_event(self, error: Exception) -> None:
...
+            tool_call=UiPathConversationToolCallEvent(
+                tool_call_id=tool_call_id,
+                executing=UiPathConversationExecutingToolCallEvent(
+                    input=tool_input,
+                ),
+            ),
+        )
+        await self.emit_message_event(executing_event)
+
     async def wait_for_resume(self) -> dict[str, Any]:
```
maxduu, line 401: I wonder if we can make the return type, our core types `UiPathConversationToolCallConfirmationEvent |  UiPathConversationToolCallEndEvent` for stronger typing?

maxduu: Ah I see, those types are BaseModels so might not be possible due to the model_dump. Could be worth a quick Claude check though, not a blocker

packages/uipath/src/uipath/_cli/_chat/_bridge.py
norman-le: Responded in the [thread](https://uipath-product.slack.com/archives/C090NH84361/p1779989231335509?thread_ts=1778561827.082409&cid=C090NH84361) but posting here too:

wait_for_resume is called after the graph suspends, but executingToolCall was sent earlier during streaming (during the mapper). The client receives it, executes, and could send endToolCall back before wait_for_resume is called. The guard checks if a response already arrived and skips the wait. Without it, the response would be cleared and wait_for_resume hangs. Unlikely in practice (network latency helps), but technically possible

maxduu: Gotcha - cool, I'm good for this but I think let's let @JoshParkSJ give final approval on this repo's PR since it mainly touches tool-confirmation logic

JoshParkSJ (approved):

packages/uipath/src/uipath/_cli/_chat/_bridge.py
JoshParkSJ: could you add this as a comment? just so future readers can also understand that it's for case where there's no tool confirmation and client side tools end fast

PR #1667, opened 2026-05-20
feat(agent): add tool-output and V2 escalation recipient types [ACTN-10480]

kunaluipath:
Adds the agent + platform models needed for the V2 HITL escalation recipient flow consumed by the langchain runtime.

uipath/agent/models:
- New literal recipient classes for Workload (9), RoundRobin (10), and CustomAssignees (11) matching the storage schema v50 surface.
- New ToolOutputRecipient class accepting types USER_ID/GROUP_ID/WORKLOAD/ ROUND_ROBIN/CUSTOM_ASSIGNEES with source="toolOutput", toolName, outputPath fields for runtime-resolved assignees.
- ToolOutputRecipient listed first in the recipient Union so payloads carrying `source` match before falling through to literal variants.

uipath/platform/action_center:
- TaskRecipientType gains Workload and RoundRobin members.
- TaskRecipient gains an optional `values` list field for the multi-assignee assigneeNamesOrEmails payload.
- _tasks_service updated to forward the new fields when present.

Backwards-compatible: existing payloads without source/values parse and serialize identically.

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/action_center/_tasks_service.py
+++ b/packages/uipath-platform/src/uipath/platform/action_center/_tasks_service.py
@@ -305,6 +305,34 @@ async def _assign_task_spec(
                     }
                 ]
             }
+        elif task_recipient.type == TaskRecipientType.WORKLOAD:
+            # This branch covers BOTH agent-side Workload criteria (single
+            # group, distributed by workload) AND agent-side CustomAssignees
+            # criteria (explicit email list — already resolved into
+            # `task_recipient.values` upstream). Both submit to the Action
+            # Center API as a "Workload" assignment; the difference is whether
+            # `values` carries one group or N emails.
+            request_spec.json = {
+                "taskAssignments": [
+                    {
+                        "taskId": task_key,
+                        "assignmentCriteria": "Workload",
+                        "assigneeNamesOrEmails": task_recipient.values
+                        or [recipient_value],
+                    }
+                ]
+            }
+        elif task_recipient.type == TaskRecipientType.ROUND_ROBIN:
+            request_spec.json = {
+                "taskAssignments": [
+                    {
+                        "taskId": task_key,
+                        "assignmentCriteria": "RoundRobin",
+                        "assigneeNamesOrEmails": task_recipient.values
+                        or [recipient_value],
+                    }
+                ]
+            }
         else:
             request_spec.json = {
                 "taskAssignments": [
--- a/packages/uipath-platform/src/uipath/platform/action_center/tasks.py
+++ b/packages/uipath-platform/src/uipath/platform/action_center/tasks.py
@@ -22,18 +22,35 @@ class TaskRecipientType(str, enum.Enum):
     GROUP_ID = "GroupId"
     EMAIL = "UserEmail"
     GROUP_NAME = "GroupName"
+    WORKLOAD = "Workload"
+    ROUND_ROBIN = "RoundRobin"
 
 
 class TaskRecipient(BaseModel):
-    """Model representing a task recipient."""
+    """Model representing a task recipient.
+
+    `value` is the single identifier (group name, group id, user id, email, …).
+    `values` is the multi-assignee form used by Workload-with-custom-emails
+    assignments; when set it takes precedence over `value` for the
+    `assigneeNamesOrEmails` payload.
+
+    Note: there is no CustomAssignees member here on purpose. The agent-side
+    CustomAssignees criteria (AgentEscalationRecipientType.CUSTOM_ASSIGNEES,
+    type 11) is resolved to a Workload assignment with the explicit email list
+    in `values` before reaching this layer, so the Action Center
+    AssignTasks API only ever sees the existing literal types.
+    """
 
     type: Literal[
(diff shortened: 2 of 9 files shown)


packages/uipath-platform/src/uipath/platform/action_center/tasks.py
```diff
@@ -22,18 +22,29 @@ class TaskRecipientType(str, enum.Enum):
     GROUP_ID = "GroupId"
     EMAIL = "UserEmail"
     GROUP_NAME = "GroupName"
+    WORKLOAD = "Workload"
+    ROUND_ROBIN = "RoundRobin"
```
dushyant-uipath, line 26: Should CustomAssignments be required as an option here?

packages/uipath/src/uipath/agent/models/agent.py
```diff
@@ -619,14 +625,86 @@ class ArgumentGroupNameRecipient(BaseEscalationRecipient):
...
+        ToolOutputRecipient,
         StandardRecipient,
         AssetRecipient,
         ArgumentEmailRecipient,
         ArgumentGroupNameRecipient,
+        WorkloadRecipient,
+        RoundRobinRecipient,
+        CustomAssigneesRecipient,
     ],
-    Field(discriminator="type"),
```
dushyant-uipath, line 636: # Why the Order Matters

`ToolOutputRecipient` **must be listed first** and must keep:

```python
source: Literal["toolOutput"]
```

as a **required field**.

These two properties are the only things that make it structurally disjoint from the literal classes below it, which share the same `type` values (for example, both `WorkloadRecipient` and `ToolOutputRecipient` accept `type=WORKLOAD`).

## Pydantic Parsing Behavior

Pydantic evaluates `Union` members **left-to-right** and stops at the first successful match:

- Payload **with** `source` → `ToolOutputRecipient` matches, literal class is never tried.
- Payload **without** `source` → `ToolOutputRecipient` fails (`source` is required), so validation falls through to the literal classes.

## Why a Discriminator Cannot Be Used

A `Field(discriminator="type")` cannot be used because the `type` literals are **not unique** across the classes.

## Critical Invariants

The following invariants must hold, otherwise silent mis-typing can occur:

1. `ToolOutputRecipient` remains the **first** member in the `Union`.
2. `ToolOutputRecipient.source` remains a **required** `Literal["toolOutput"]`.
3. No literal class below it gains an optional `source` field.

packages/uipath-platform/src/uipath/platform/action_center/_tasks_service.py
```diff
@@ -305,6 +305,28 @@ async def _assign_task_spec(
                     }
                 ]
             }
+        elif task_recipient.type == TaskRecipientType.WORKLOAD:
```
dushyant-uipath, line 308: NIT - Can clarify here that even custom assignments will follow this criteria

dushyant-uipath (approved):

packages/uipath-platform/src/uipath/platform/action_center/tasks.py
kunaluipath: Intentionally omitted: `TaskRecipient` is the platform-API enum and the API has no `CUSTOM_ASSIGNEES` member — agent-side `CustomAssignees` resolves to `Workload` (type 9) before reaching the task-creation endpoint (see `_assign_task_spec` in `_tasks_service.py`). Added a docstring note on the enum to make this explicit.

packages/uipath/src/uipath/agent/models/agent.py
kunaluipath: Thanks, this is exactly right — kept `ToolOutputRecipient` first in the union and `source: Literal["toolOutput"]` required. Added a `Critical Invariants` comment block on `AgentEscalationRecipient` documenting the three rules (ordering, required source, no source on literal classes) so future edits don't accidentally break the discriminator.

packages/uipath-platform/src/uipath/platform/action_center/_tasks_service.py
kunaluipath: Good call — added an inline comment in the `WORKLOAD` branch of `_assign_task_spec` clarifying that this branch handles both agent-`Workload` recipients and agent-`CustomAssignees` (which map to `Workload` server-side).

kunaluipath:
Rebased on `main` and merge-ready.

- Resolved `packages/uipath/pyproject.toml` and `packages/uipath-platform/pyproject.toml` conflicts by taking the newer main versions and bumping forward: `uipath==2.11.1`, `uipath-platform==0.1.68`.
- Regenerated both `uv.lock` files.
- Local tests pass: uipath 1972 ✓, uipath-platform 1239 ✓ (+ 7 skipped).
- All required checks green: `commit-lint` ✓, `lint` ✓, `test` ✓. The 3 failing checks (`debug-breakpoints`, `dev-console`, `template-agent`) are non-required langchain cross-tests that fail due to PyPI lag — once this PR merges and publishes 2.11.1, the companion uipath-langchain PR https://github.com/UiPath/uipath-langchain-python/pull/867 can rebase to pick up the new types and its SonarCloud Quality Gate will pass.

PR #1687, opened 2026-05-27
feat: send jobkey to ECS [ECS-1819]

gcuip:


packages/uipath-platform/src/uipath/platform/common/_job_context.py
```diff
@@ -0,0 +1,16 @@
+from os import environ as env
+
+from .constants import ENV_JOB_KEY, HEADER_JOB_KEY
+
+
+def header_job_key() -> dict[str, str]:
```
radu-mocanu, line 5: let s cache this property and use `UiPathConfig.job_key` instead

radu-mocanu (approved):

PR #1692, opened 2026-05-29
fix(eval): use function calling for tool/input mocking so non-OpenAI models work (AE-1646)

Chibionos:
## Summary

Tool simulation and input generation in **Studio Debug** and **Evaluation Set** runs failed with `AGENT_RUNTIME.UNEXPECTED_ERROR` for **non-OpenAI models** (Anthropic Claude via Bedrock, Gemini), but worked with OpenAI/GPT.

Fixes **AE-1646** (customer: Sarasota Memorial Health Care System).

## Root cause (live-verified on alpha)

Both eval mockers requested structured output via OpenAI-only `response_format` json_schema and parsed `response.choices[0].message.content`. The normalized LLM Gateway only honors `response_format` for OpenAI models — and the failure shape **differs per provider** (verified by live calls, not assumed):

| Provider | `response_format` behavior on the gateway |
|---|---|
| OpenAI (gpt-4.1-mini) | honored — valid JSON content, native `$defs` support |
| Claude (sonnet-4-5) | ignored — returns **plain prose content** (e.g. `'Tokyo'`) |
| Gemini (2.5-pro) | returns **empty content** |

Both non-OpenAI shapes broke `json.loads(...)` → wrapped as `UiPathMockResponseGenerationError` → `AGENT_RUNTIME.UNEXPECTED_ERROR`. Note the prose case matters: an earlier revision of this PR fell back only on *empty* content, which fixed Gemini but left Claude — the customer's model — still failing. The live e2e below caught that.

Regression from #1555, which started routing the **agent's** model into simulations; before that, simulation always used a fixed OpenAI model.

## Fix

Provider-aware strategy classes in `eval/mocks/_structured_output.py` (mirroring `llm_as_judge_evaluator`, whose docstring already states function calling is the cross-provider way to get structured output):

- **`OpenAIStructuredOutput`** — prefers `response_format` (reliable for OpenAI incl. `$defs`), falls back to a forced tool call on empty/non-JSON content or request error.
- **`AnthropicStructuredOutput` / `GeminiStructuredOutput`** — go **straight to a forced tool call** (their `response_format` is known-broken on the gateway; no wasted request per simulation).
- **Unknown providers** — `response_format` first with tool fallback (safe default).

The forced tool wraps the output/input schema under a `response` property (`tool_choice=required`), and reads `tool_calls[0].arguments["response"]` (already a parsed dict). `$defs`/`$ref` are **inlined** so tool parameters are self-contained (the gateway accepts `$defs` in `response_format` but not in tool parameters); sibling keys on `$ref` nodes (e.g. field descriptions) are merged over the inlined definition so LLM guidance survives. `chat_completions` now accepts **raw-dict tools** (pass-through) so arbitrary nested schemas reach the gateway verbatim.

## E2E validation (live against alpha gateway)

Full customer path per model — `@mockable` → `Mocker.response` → `generate_structured_output` → **real gateway** → Pydantic coercion of the result. Tool simulation used a nested return model (enum + list of sub-models ⇒ real `$defs`/`$ref`); input generation ran `generate_llm_input` the same way.

| Model | Tool simulation (nested `$defs`) | Input generation | Calls |
|---|---|---|---|
| gpt-4.1-mini-2025-04-14 | ✅ coerced to `Calculation`, total=42.0 | ✅ | 1 (`response_format`) |
| anthropic.claude-sonnet-4-5-20250929-v1:0 | ✅ coerced to `Calculation`, total=42.0 | ✅ | 1 (tool call) |
| gemini-2.5-pro | ✅ coerced to `Calculation`, total=42.0 | ✅ | 1 (tool call) |

Negative control: reverting the non-JSON-content fallback and re-running the Claude e2e reproduces the exact customer error (`UiPathMockResponseGenerationError: Expecting value: line 1 column 1 (char 0)`), confirming the e2e discriminates.

## Tests

- Strategy routing: Claude/Gemini → single forced tool call, no `response_format`; OpenAI → `response_format` preferred; unknown → fallback chain (incl. prose-content, empty-content, and request-error triggers).
- `$defs` inlining: self-contained output, sibling-key merge, cyclic-ref handling, caller schema not mutated.
- `test_raw_dict_tool_passthrough_mocked` (platform): nested array schema forwarded byte-for-byte.
- Full `tests/cli/eval` suite (405) + platform mocked LLM tests green; ruff + mypy clean on both packages.

## Review feedback addressed

- **@mjnovice** — separate per-provider classes: done (`OpenAIStructuredOutput`, `AnthropicStructuredOutput`, `GeminiStructuredOutput`, shared `ToolCallStructuredOutput` base).
- **Copilot** — description/implementation mismatch: resolved by the provider strategies (this description now matches); `chat_completions` docstring updated for raw-dict tools.

### DIFF
--- a/packages/uipath/src/uipath/eval/mocks/_structured_output.py
+++ b/packages/uipath/src/uipath/eval/mocks/_structured_output.py
@@ -12,6 +12,50 @@ from typing import Any
 
 RESPONSE_TOOL_NAME = "submit_tool_response"
 RESPONSE_KEY = "response"
+_DEFS_PREFIX = "#/$defs/"
+
+
+def _inline_defs(
+    schema: dict[str, Any],
+) -> tuple[dict[str, Any], dict[str, Any]]:
+    """Inline ``$defs``/``$ref`` into a self-contained schema.
+
+    Nested Pydantic models and enums emit root ``$defs`` referenced by ``$ref``.
+    The normalized gateway accepts those in ``response_format`` but not inside a
+    tool's ``parameters``, so they are inlined here. Self-referential definitions
+    cannot be inlined without looping; any ``$ref`` reached while its target is
+    already on the current resolution path is left untouched and its definitions
+    are returned so the caller can keep them reachable.
+
+    Returns:
+        A tuple of (inlined schema, leftover ``$defs`` needed for cyclic refs).
+    """
+    defs = schema.get("$defs", {})
+    leftover: dict[str, Any] = {}
+
+    def resolve(node: Any, active: frozenset[str]) -> Any:
+        if isinstance(node, dict):
+            ref = node.get("$ref")
+            if isinstance(ref, str) and ref.startswith(_DEFS_PREFIX):
+                name = ref[len(_DEFS_PREFIX) :]
+                if name in defs and name not in active:
+                    return resolve(defs[name], active | {name})
+                # Cyclic or unknown ref: keep it and preserve its definition.
+                if name in defs:
+                    leftover[name] = defs[name]
+                return dict(node)
+            return {
+                key: resolve(value, active)
+                for key, value in node.items()
+                if key != "$defs"
+            }
+        if isinstance(node, list):
+            return [resolve(item, active) for item in node]
+        return node
+
+    root = {key: value for key, value in schema.items() if key != "$defs"}
+    inlined = resolve(root, frozenset())
+    return inlined, leftover
 
 
 def build_response_tool(schema: dict[str, Any], description: str) -> dict[str, Any]:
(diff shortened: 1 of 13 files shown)


Chibionos:
Re: the `uipath-langchain` cross-test heads-up above — that was from an earlier commit. After the adaptive fix (prefer `response_format`, fall back to function calling only when content is empty), the cross-tests pass on the latest commit: `langchain-cross / {alpha,cloud,staging}` and `test-uipath-langchain` (3.11/3.12/3.13 × ubuntu/windows) are all green. The heads-up comment is stale and can be disregarded.

Full suite is green: all eval test-cases (calculator-evals, simulation-testcase, tools-evals, etc.) pass across alpha/cloud/staging; lint, SonarCloud, build all pass.

packages/uipath/src/uipath/eval/mocks/_structured_output.py
```diff
@@ -0,0 +1,168 @@
...
+from uipath.platform.chat.llm_gateway import RequiredToolChoice
+
+RESPONSE_TOOL_NAME = "submit_tool_response"
+RESPONSE_KEY = "response"
+_DEFS_PREFIX = "#/$defs/"
+
+logger = logging.getLogger(__name__)
+
+
+def _inline_defs(
```
mjnovice, line 31: Can we have separate classes for how we are doing Anthropic, OpenAI, Gemini etc. ?

mjnovice (approved):
Minor comment about making the generate_structured_output more modular.

packages/uipath/src/uipath/eval/mocks/_structured_output.py
Chibionos: Done in a8a33078 — split into per-provider strategy classes: `OpenAIStructuredOutput` (response_format preferred, tool fallback), `AnthropicStructuredOutput` and `GeminiStructuredOutput` (straight to a forced tool call, since their response_format is broken on the gateway — live-verified: Claude returns prose, Gemini empty content), with a shared `ToolCallStructuredOutput` base. Also avoids the wasted known-dead request per simulation for non-OpenAI models. E2E results on all 3 models are in the updated PR description.

PR #1695, opened 2026-05-29
feat: add stable id to uipath.json for packaging and spans

radu-mocanu:
## Summary
- `uipath init` mints a stable `id` (GUID) into `uipath.json`, preserved across re-init and backfilled for existing projects
- `uipath pack` sources the package `projectId` from `uipath.json#id`, validating it is a GUID (and failing otherwise), falling back to the legacy `.telemetry.json` `ProjectKey` / SW project id
- llmops spans, trace baggage, and eval telemetry resolve the agent id through one source: `uipath.json#id` first, then the legacy env vars (`UIPATH_AGENT_ID` / `UIPATH_PROJECT_ID` / `PROJECT_KEY`) for already-deployed agents; a non-GUID `id` is ignored so it can never reject span ingestion
- `uipath init` no longer writes `.uipath/.telemetry.json`; an existing one is still honored for the id and telemetry attribution
- the `id` field is documented in `uipath.schema.json` / `uipath.spec.md`
- allow `langchain-cross` tests to run against 'uncompatible' libraries from PR (similar to https://github.com/UiPath/uipath-python/pull/1723)

## Why

a coded agent or function needs one stable identifier that survives authoring, packaging, and runtime so traces and deployed releases correlate back to the same project. anchoring it in `uipath.json` decouples the id from telemetry being enabled, and routing spans, baggage and telemetry through a single resolver keeps them from diverging.

### uipath

```toml
[project]
dependencies = [
  "uipath==2.11.0.dev1016956769",

  # "uipath>=2.11.0.dev1016950000,<2.11.0.dev1016960000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = ["uipath-platform==0.1.67.dev1016956769"]
```

### uipath-platform

```toml
[project]
dependencies = [
  "uipath-platform==0.1.67.dev1016956769",

  # "uipath-platform>=0.1.67.dev1016950000,<0.1.67.dev1016960000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/common/_span_utils.py
+++ b/packages/uipath-platform/src/uipath/platform/common/_span_utils.py
@@ -2,9 +2,11 @@ import inspect
 import json
 import logging
 import os
+import uuid
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import IntEnum
+from functools import lru_cache
 from os import environ as env
 from typing import Any, Dict, List, Optional
 
@@ -17,6 +19,40 @@ logger = logging.getLogger(__name__)
 
 # SourceEnum.CodedAgents = 10 (default for Python SDK / coded agents)
 DEFAULT_SOURCE = 10
+
+
+@lru_cache(maxsize=1)
+def _read_config_id() -> str | None:
+    """Return a valid GUID ``id`` from ``uipath.json``, cached for the process lifetime."""
+    from uipath.platform.common._config import UiPathConfig
+
+    try:
+        config_file = json.loads(UiPathConfig.config_file_path.read_text())
+    except (FileNotFoundError, json.JSONDecodeError):
+        return None
+
+    project_id = config_file.get("id")
+    if not isinstance(project_id, str):
+        logger.warning("'id' field not present in uipath.json")
+        return None
+
+    try:
+        uuid.UUID(project_id)
+    except ValueError:
+        logger.warning("Ignoring uipath.json 'id' %r: not a valid GUID.", project_id)
+        return None
+
+    return project_id
+
+
+def resolve_project_id() -> str | None:
+    """Resolve the project id.
+
+    Prefers ``uipath.json#id``, falls back to env vars.
+    """
+    from uipath.platform.common._config import UiPathConfig
+
+    return _read_config_id() or UiPathConfig.agent_id or UiPathConfig.project_key
 
 
 class AttachmentProvider(IntEnum):
@@ -281,9 +317,11 @@ class _SpanUtils:
             ]
             attributes_dict["links"] = links_list
 
+        if agent_id := resolve_project_id():
+            attributes_dict["agentId"] = agent_id
+
         # Add process context attributes from environment variables
         for env_key, attr_key in (
-            ("PROJECT_KEY", "agentId"),
             ("UIPATH_PROCESS_KEY", "agentName"),
             ("UIPATH_PROCESS_VERSION", "agentVersion"),
         ):
@@ -297,10 +335,8 @@ class _SpanUtils:
         # Top-level fields for internal tracing schema
         execution_type = attributes_dict.get("executionType")
         agent_version = attributes_dict.get("agentVersion")
-        reference_id = (
-            env.get("UIPATH_AGENT_ID")
-            or attributes_dict.get("agentId")
-            or attributes_dict.get("referenceId")
+        reference_id = attributes_dict.get("agentId") or attributes_dict.get(
+            "referenceId"
         )
         verbosity_level = attributes_dict.get("verbosityLevel")
 
--- a/packages/uipath/src/uipath/_cli/cli_init.py
+++ b/packages/uipath/src/uipath/_cli/cli_init.py
@@ -9,6 +9,7 @@ import uuid
 from pathlib import Path
 from typing import Any
 
+import anyio
 import click
 from graphtty import RenderOptions, render
 from graphtty.themes import TOKYO_NIGHT
@@ -30,13 +31,11 @@ from uipath.runtime import (
 )
 from uipath.runtime.schema import UiPathRuntimeGraph, UiPathRuntimeSchema
 
-from .._utils.constants import ENV_TELEMETRY_ENABLED
-from ..telemetry._constants import _PROJECT_KEY, _TELEMETRY_CONFIG_FILE
 from ._telemetry import track_command
 from ._utils._common import determine_project_type
 from ._utils._console import ConsoleLogger
 from ._utils._constants import AGENT_INITIAL_CODE_VERSION, SCHEMA_VERSION
-from ._utils._project_files import read_toml_project
+from ._utils._project_files import read_toml_project, resolve_existing_project_id
 from .middlewares import Middlewares
 from .models.runtime_schema import Bindings, EntryPoint
 from .models.uipath_json_schema import UiPathJsonConfig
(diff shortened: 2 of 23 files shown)


packages/uipath-platform/src/uipath/platform/common/_span_utils.py
```diff
@@ -281,9 +301,14 @@ def otel_span_to_uipath_span(
             ]
             attributes_dict["links"] = links_list
 
+        # agentId: prefer the stable id from uipath.json (cached), falling back
+        # to the legacy PROJECT_KEY env var injected by the executor.
+        agent_id = _read_config_agent_id() or env.get("PROJECT_KEY")
```
RobinMennens-UiPath: We now resolve `agentId` here from `uipath.json` → `PROJECT_KEY`. But `llm_trace_context.py:43` sets the baggage `agentId` from `UiPathConfig.agent_id` (`UIPATH_AGENT_ID` env / `project_id`), and `reference_id` just below (line 325) starts from `UIPATH_AGENT_ID` env again. So for one running agent the span attribute, the trace baggage and the referenceId can all resolve to different values. Can we funnel these through a single source of truth?

RobinMennens-UiPath: nit: `env.get("PROJECT_KEY")` — we already have `ENV_PROJECT_KEY = "PROJECT_KEY"` in `constants.py`. Since you rewrote this line, might as well use it.

packages/uipath/src/uipath/_cli/cli_init.py
```diff
@@ -429,10 +429,22 @@ async def initialize() -> list[UiPathRuntimeSchema]:
...
                     config = UiPathJsonConfig.create_default()
+                    config.agent_id = resolve_existing_project_id(
+                        current_directory
+                    ) or str(uuid.uuid4())
                     config.save_to_file(config_path)
                     console.success(f"{Action.CREATED.value} '{config_path}' file.")
                 else:
-                    console.info(f"'{config_path}' already exists, skipping.")
+                    # backfill agentId if not present
+                    config = UiPathJsonConfig.load_from_file(str(config_path))
```
RobinMennens-UiPath: The backfill does `load_from_file` → set agent_id → `save_to_file`, which re-serializes the whole model. Because `to_json_string` only does `exclude_none`, any defaulted fields the user had omitted (`packOptions`, `agents`, `functions`) get materialized, and extra top-level keys get pushed to the end. So we rewrite a richer/reordered file than the user had just to add one field. Intended? If we only want to add `agentId` a targeted write would be less surprising. The backfill test only asserts `functions` survives, so this widening isn't covered.

RobinMennens-UiPath (commented):
Took a pass focused on the agentId behavior across the coded agent types, regressions, and the spans. Two things I'd like to align on before this goes in:

For plain functions and declarative agents this is end-to-end correct. But langgraph/llamaindex agents build their own `operate.json` in the plugin repos and won't source `projectId` from `uipath.json#agentId` until we mirror this there. Is there a follow-up tracking that, or is the expectation that those plugins start reading `uipath.json#agentId`?

Separately, on alignment with the `vertical-solutions-check-agent-id.yml` check: that check looks for a coded agent's ID only in `.uipath/.telemetry.json#ProjectKey` or `.uipath/local-project-id.json#projectId` — not in `uipath.json#agentId`. With telemetry disabled, `create_telemetry_config_file` doesn't write `.telemetry.json`, and we don't write `local-project-id.json` either, so an agent inited by this PR would have its ID only in `uipath.json` and the CI check would report it as missing. Are we updating that check to read `uipath.json#agentId` too, or should `init` also write one of the files it expects? These two need to agree.

radu-mocanu:
> But langgraph/llamaindex agents build their own operate.json in the plugin repos and won't source projectId from uipath.json#agentId

This is not true.
All UiPath Coded Agents are using the [same packager](https://github.com/UiPath/uipath-python/blob/main/packages/uipath/src/uipath/_cli/cli_pack.py). No changes are needed in the integration repos.

radu-mocanu:
> Separately, on alignment with the vertical-solutions-check-agent-id.yml check

This is not a concern for the coded agents SDK. the changes are backwards compatible, meaning that existing IDs will still be honored. newly created agents will indeed drop `telemetry.json` file. Any consumer (like that gh action) need to be updated to reflect the new behaviour -> `uipath.json` as the source of truth.

packages/uipath-platform/src/uipath/platform/common/_span_utils.py
radu-mocanu: the `span's AgentId` attribute uses `uipath.json#Id` as the first candidate https://github.com/UiPath/uipath-python/pull/1695/changes#diff-92971d037c48f96ec23800bcb96b8016a1906d8c0ae422bb5e55e08830bcd9e3R308.

packages/uipath/src/uipath/_cli/cli_init.py
radu-mocanu: good catch

alexandra-c (changes requested):
Could we extend this to cover function project types as well? `agentId` isn't the right name for a coded function, could we detect the project type and use `functionId` accordingly, or pick a generic name that works for both?

RobinMennens-UiPath:
> Could we extend this to cover function project types as well? `agentId` isn't the right name for a coded function, could we detect the project type and use `functionId` accordingly, or pick a generic name that works for both?

If you ask me, **everything** that can run on the UiPath platform should have a unique GUID identifier that never changes. RPA, coded functions, coded agents, low code agents, etc.
`ProjectId` kind of already seems a generic candidate for that.. but `agentId` already got introduced as a special ID just for agents. 
We either keep introducing special IDs such as `functionId`, or we try to limit the number of IDs we have and stick to a generic `ProjectId` or something similar
CC @JosephMar

RobinMennens-UiPath (approved):

radu-mocanu:
> We either keep introducing special IDs such as functionId, or we try to limit the number of IDs we have and stick to a generic ProjectId or something similar

Agreed. the problem is that we already have a ProjectKey (similar to project id, that is the release key from orchestrator). to be consistent with the agentId naming (which is already the established field for low code agents today) I would use functionId for functions. I will also consult with the team regarding this since it would be a spec decision

JosephMar:
> > We either keep introducing special IDs such as functionId, or we try to limit the number of IDs we have and stick to a generic ProjectId or something similar
> 
> Agreed. the problem is that we already have a ProjectKey (similar to project id, that is the release key from orchestrator). to be consistent with the agentId naming (which is already the established field for low code agents today) I would use functionId for functions. I will also consult with the team regarding this since it would be a spec decision

IMO every product should define its own GUID (I imagine you'll need this exposed downstream at some point). For Agents we are exposing a concrete AgentIId which must be separate from ProjectId (because of flow) so the generic approach might not work well unless you also include a discriminator, but at that point having separate IDs is probably similar overhead

@jepadil23 FYI this is similar to what we are solving for Trace service ID hierarchy

danflorescu96 (approved):

jepadil23 (approved):
idea LGTM, only resolve if possible build issues

radu-mocanu:
Settled on `uipath.json#Id` for both agents and functions. Updated also the spec docs.

ctiliescuuipath (approved):

PR #1696, opened 2026-06-02
fix: eph folder key issue

CalebMartinUiPath:
make sure the create ephemeral index call has folder key. follows the same path to get it as our other methods now which also checks env vars

relevant issue: https://uipath-product.slack.com/archives/C060F2VLC20/p1780340678652109?thread_ts=1779989349.104999&cid=C060F2VLC20

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
+++ b/packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
@@ -2262,13 +2262,11 @@ class ContextGroundingService(FolderContext, BaseService):
         folder_key: Optional[str] = None,
         folder_path: Optional[str] = None,
     ) -> RequestSpec:
-        folder_key = self._resolve_folder_key(folder_key, folder_path)
-
+        # Folder key not needed by retrieve by id. Adding it breaks ephemeral indexes
         return RequestSpec(
             method="GET",
             endpoint=Endpoint(f"/ecs_/v2/indexes/{id}"),
             headers={
-                **header_folder(folder_key, None),
                 **header_job_key(),
             },
         )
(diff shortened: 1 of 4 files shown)


packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
```diff
@@ -696,8 +696,7 @@ def create_ephemeral_index(
         Returns:
             ContextGroundingIndex: The created index information.
         """
-        if folder_key is not None or folder_path is not None:
-            folder_key = self._resolve_folder_key(folder_key, folder_path)
+        folder_key = self._resolve_folder_key(folder_key, folder_path)
```
gcuip, line 699: not sure I understand this change entirely. we should resolve the folder_key only when either the folder_key / folder_path are passed. otherwise, the index creation will not be folder-bound.

see https://github.com/UiPath/uipath-langchain-python/blob/main/src/uipath_langchain/agent/tools/internal_tools/deeprag_tool.py#L121

CalebMartinUiPath: here is my understanding: 

there are 3 places it can get the folder key from:
1. folder_key passed in from method params
2. folder_path passed in from method params
3. env var (from self._folder_key) https://github.com/UiPath/uipath-python/blob/c736c3a9b6b1aaf70283889bb95f3d116612118b/packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py#L2534


with that if statement we are not allowing it to check #3.

in all other places in the sdk we directly call _resolve_folder_key so I don't understand what is unique about ephemeral indexes that we need to wrap it an if statement.

I don't know if this inconsistency is what is causing the customer issue, but it is my best guess. and I don't see what changing it will hurt.

JeremyReist2: when is the env var set? I think we can check if we got the folder in the request

PR #1701, opened 2026-06-04
 move pii masking service to llmops

yashwagle1:


packages/uipath-platform/src/uipath/platform/pii_detection/pii_utilities.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    entities: list[PiiEntity] = []
+    for doc in response.response:
+        entities.extend(doc.pii_entities)
+    for file in response.files:
+        entities.extend(file.pii_entities)
+    return rehydrate_from_pii_entities(masked_text, entities)
+
+
+def _add_escape_characters(text: str) -> str:
```
saksharthakkar: I know it's a copy of the existing implementation and needed in C#, but this is not necessary for python and can add unnecessary characters.

saksharthakkar:
nit: title is misleading ...

saksharthakkar (approved):
approved with couple of nit comments

PR #1705, opened 2026-06-09
feat(resource-catalog): add ENTITY to ResourceType enum

msbhavana:
## Summary
- Adds `ENTITY = "entity"` to the `ResourceType` enum in `resource_catalog.py`
- Enables entity bindings to go through the resource catalog lookup during `uipath push`
- When an entity is found in the catalog, it is created as a **referenced resource** instead of being skipped or falling back to a virtual resource
- Updates existing test to use `choiceSet` as the unsupported kind example (instead of `entity`)
- Adds new test `test_entity_binding_catalog_hit_creates_reference` to verify entities are properly referenced

## Context
Entity bindings in `bindings.json` were previously skipped during push because the `ResourceType` enum had no `ENTITY` member. This caused `_find_in_resource_catalog` to short-circuit (returning `None`), and since `entity` is also not in the `supported_virtual_kinds` set, the binding was silently skipped.

With this change, folder-level entities declared in bindings can be properly resolved and linked into Studio Web solutions as referenced resources.

## Test plan
- [x] Verified locally: `uip codedagent push` with entity binding creates a referenced resource (`kind = Entity, type = Native`)
- [ ] Existing tests updated to reflect new behavior
- [ ] New test added for entity catalog hit → reference creation

### DIFF
--- a/packages/uipath-platform/uv.lock
+++ b/packages/uipath-platform/uv.lock
@@ -3,7 +3,7 @@ revision = 3
 requires-python = ">=3.11"
 
 [options]
-exclude-newer = "2026-06-06T13:38:31.678016Z"
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
 exclude-newer-span = "P2D"
 
 [options.exclude-newer-package]
@@ -1095,7 +1095,7 @@ dev = [
 
 [[package]]
 name = "uipath-platform"
-version = "0.1.61"
+version = "0.1.62"
 source = { editable = "." }
 dependencies = [
     { name = "httpx" },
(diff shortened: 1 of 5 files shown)


sandeeprao-uipath (approved):

packages/uipath-platform/uv.lock
```diff
@@ -3,7 +3,7 @@ revision = 3
 requires-python = ">=3.11"
 
 [options]
-exclude-newer = "2026-06-06T13:38:31.678016Z"
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
```
cosmyo, line 6: please revert this ... add current date if necessary

PR #1675, opened 2026-05-22
docs(studio-web): document coded function integration [PRODEV-605]

alexandra-c:
Extend the Studio Web integration guide to cover coded functions alongside coded agents

[PRODEV-605](https://uipath.atlassian.net/browse/PRODEV-605)

[PRODEV-605]: https://uipath.atlassian.net/browse/PRODEV-605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1706, opened 2026-06-09
docs(core): correct project-type marker and I/O typing for functions/…

AlexBizon:
…agents

The functions/agents guides (added in #1675) claim a `[tool.uipath] type = "function"|"agent"` marker in pyproject.toml is required to identify the project type. It isn't — `uipath new` scaffolds pyproject without it and no sample carries it. The project type is determined by the `functions` / `agents` map in `uipath.json`. Remove the marker from the examples and the "is required" claims.

Also broaden the I/O guidance: Input/Output can be a stdlib @dataclass, a pydantic BaseModel, or pydantic.dataclasses.dataclass (the samples use pydantic), and the entry point may be sync `def` or `async def`.

### DIFF
--- a/packages/uipath/docs/core/agents.md
+++ b/packages/uipath/docs/core/agents.md
@@ -91,22 +91,24 @@ The example below uses LangChain. Swap `uipath-langchain` for the framework of y
 
 ```
 my-agent/
-├── main.py            # agent logic
+├── main.py            # agent graph
+├── langgraph.json     # graph entry points (framework-specific)
 ├── pyproject.toml     # project metadata and dependencies
-├── uipath.json        # entry point declarations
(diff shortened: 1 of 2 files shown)


radu-mocanu (approved):

packages/uipath/docs/core/agents.md
```diff
@@ -118,18 +120,15 @@ description = "..."
 authors = [{ name = "Your Name", email = "you@example.com" }]
 requires-python = ">=3.11"
 dependencies = ["uipath>=2.0", "uipath-langchain>=2.0"]
```
radu-mocanu, line 122: nit: `uipath` is a dependency of `uipath-langchain`, no need to declare it in the dependencies list here

PR #1707, opened 2026-06-10
chore: consolidate uipath._utils.constants into uipath-platform.common.constants

### LINKED ISSUE #1415: UiPath general constants duplicated
There should be a single source of truth for constants; they are duplicated currently in:
1. https://github.com/UiPath/uipath-python/blob/main/packages/uipath/src/uipath/_utils/constants.py
2. https://github.com/UiPath/uipath-python/blob/main/packages/uipath-platform/src/uipath/platform/common/constants.py

vldcmp-uipath:
Fixes https://github.com/UiPath/uipath-python/issues/1415

## Summary

* `uipath/src/uipath/_utils/constants.py`
* `uipath-platform/src/uipath/platform/common/constants.py` 

have drifted into two near-identical modules. External packages (`uipath-llm-client-python`, `uipath-langchain-python`) import from `uipath.platform.common.constants`; internal `uipath` consumers import from `uipath._utils.constants`. A silent value drift between the two would create hard-to-debug cross-package bugs — same name, different value.

This PR consolidates onto a single source of truth in `uipath-platform`, leaving a thin shim with a deprecation warning at the old path for backward compatibility.

**Version bumps:** uipath-platform 0.1.86 → 0.1.87, uipath 2.12.1 → 2.12.2.

## Follow-ups (not in this PR)
- After enough release cycles, the shim can be deleted outright.

### DIFF
--- a/packages/uipath-platform/pyproject.toml
+++ b/packages/uipath-platform/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-platform"
-version = "0.1.86"
+version = "0.1.87"
 description = "HTTP client library for programmatic access to UiPath Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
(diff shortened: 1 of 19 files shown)


radu-mocanu (changes requested):
we shouldn't import `uipath.platform.common.constants` from the global CLI entrypoint. we only need the `DOTENV_FILE`, but importing that module path executes `uipath.platform.__init__` and `uipath.platform.common.__init__`, which pull in a large part of the platform SDK.
this will add latency to CLI commands that do not use platform library (eg `--help` and `--version`).

I would either add the import under `load_environment_variables` (and avoid calling the function globaly 
```
  @click.group(...)
  def cli(...):
      load_environment_variables()
      add_cwd_to_path()
```
). 
or just use ".env" literal.

let s go with the second one to keep this PR scoped, we can make an improvement later

radu-mocanu (approved):

PR #1720, opened 2026-06-15
feat: optionally emit end exchange event to CAS [JAR-9933]

andrewwan-uipath:
- Honor the endExchange fpsProperty in conversational runs
- Passes end_exchange=ctx.end_exchange into UiPathChatRuntime at both construction sites (cli_run.py, cli_debug.py), so the conversationalService.endExchange property added in uipath-runtime 0.11.2 actually takes
  effect. When false, the exchange-end event is skipped and the exchange stays open for a downstream consumer.
- related to this PR: https://github.com/UiPath/uipath-runtime-python/pull/127


maxduu (approved):
Approved but also was leaning the original approach as well, maybe let's see @cristipufu 's response to this [thread](https://github.com/UiPath/uipath-runtime-python/pull/127#discussion_r3431527283) first

PR #1724, opened 2026-06-16
fix(eval): skip synthesized tool spans in per-call extractors

ajay-kesavan:
## Summary
- Add `_real_tool_name()` predicate that skips spans carrying `tool.synthesized=True`.
- Route `extract_tool_calls_names`, `extract_tool_calls`, `extract_tool_calls_outputs` through it.
- `trace_to_str` is intentionally **unchanged** so trajectory evaluators continue to see the synthesized BPMN container context inside their LLM-judge prompt.

## Why
`Agents/python-eval-worker/workflows/eval/flow_trace.py:flow_spans_to_readable` synthesizes `tool.name` on BPMN `ProcessRun` (whole .flow) and `ElementRun` (e.g. ServiceTask) spans so `trace_to_str()` can render them. The four per-call evaluators (Tool Call Count / Order / Args / Output) share the same `tool.name` filter and end up treating those container spans as tool invocations — so the "Actual" column in a flow-inline-agent run shows `FlowExecution`, `ServiceTask: Autonomous Agent`, and the real tool (`webSearch1`), instead of just the real tool.

This PR closes the gate on the per-call side. The companion PR in `UiPath/Agents` stamps `tool.synthesized=True` on the synthesized attributes; until that lands this is a no-op (nothing in the wild sets the attribute today).

## Type / signature impact
None. `ReadableSpan.attributes` is an open OTel dict; the new key is purely additive. `ToolCall`, `ToolOutput`, all four evaluator config/criteria/justification classes, and every extractor signature are unchanged.

## Test plan
- [x] `pytest tests/evaluators/test_evaluator_helpers.py` — 52 pass, including new `test_extractors_skip_synthesized_tool_spans`
- [x] `pytest tests/cli/eval/` — 437 pass
- [x] `ruff check` + `mypy` on the modified file — clean

## Companion PR
- UiPath/Agents#5584 — stamps `tool.synthesized=True` in `flow_spans_to_readable`.

### DIFF
--- a/packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
+++ b/packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
@@ -29,7 +29,11 @@ COMMUNITY_agents_SUFFIX = "-community-agents"
 def _real_tool_attrs(span: ReadableSpan) -> Mapping[str, Any] | None:
     """Return span.attributes if this is a real tool invocation, else None."""
     attrs = span.attributes
-    if not attrs or attrs.get("tool.synthesized", False) or not attrs.get(TOOL_NAME_ATTR):
+    if (
+        not attrs
+        or attrs.get("tool.synthesized", False)
+        or not attrs.get(TOOL_NAME_ATTR)
+    ):
         return None
     return attrs
 
--- a/packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
+++ b/packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
@@ -26,7 +26,7 @@ COMPARATOR_MAPPINGS = {
 COMMUNITY_agents_SUFFIX = "-community-agents"
 
 
-def _real_tool_attrs(span: ReadableSpan) -> Mapping[str, Any] | None:
+def _unsynthesized_tool_attrs(span: ReadableSpan) -> Mapping[str, Any] | None:
     """Return span.attributes if this is a real tool invocation, else None."""
     attrs = span.attributes
     if (
@@ -50,7 +50,7 @@ def extract_tool_calls_names(spans: Sequence[ReadableSpan]) -> list[str]:
     tool_calls_names = []
 
     for span in spans:
-        if (attrs := _real_tool_attrs(span)) is not None:
+        if (attrs := _unsynthesized_tool_attrs(span)) is not None:
             tool_calls_names.append(str(attrs[TOOL_NAME_ATTR]))
 
     return tool_calls_names
@@ -68,7 +68,7 @@ def extract_tool_calls(spans: Sequence[ReadableSpan]) -> list[ToolCall]:
     tool_calls = []
 
     for span in spans:
-        if (attrs := _real_tool_attrs(span)) is not None:
+        if (attrs := _unsynthesized_tool_attrs(span)) is not None:
             tool_name = str(attrs[TOOL_NAME_ATTR])
             try:
                 input_value: Any = attrs.get("input.value", {})
@@ -99,7 +99,7 @@ def extract_tool_calls_outputs(spans: Sequence[ReadableSpan]) -> list[ToolOutput
     potential_output_keys = ["content"]
     tool_calls_outputs = []
     for span in spans:
-        if (attrs := _real_tool_attrs(span)) is not None:
+        if (attrs := _unsynthesized_tool_attrs(span)) is not None:
             tool_name = str(attrs[TOOL_NAME_ATTR])
             output = attrs.get("output.value", "")
             final_output = ""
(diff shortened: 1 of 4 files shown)


packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
```diff
@@ -24,6 +24,21 @@
 COMMUNITY_agents_SUFFIX = "-community-agents"
 
 
+def _real_tool_name(span: ReadableSpan) -> str | None:
+    """Return the tool name for a real tool-invocation span, else None.
+
+    Spans synthesized for trajectory rendering (e.g. BPMN ProcessRun/ElementRun
+    container spans) carry tool.name but mark themselves with tool.synthesized
+    so the per-call extractors below can skip them.
```
ajay-kesavan: remove this and give a one line comment only

packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
```diff
@@ -55,20 +69,20 @@ def extract_tool_calls(spans: Sequence[ReadableSpan]) -> list[ToolCall]:
     tool_calls = []
 
     for span in spans:
-        if span.attributes and (tool_name := span.attributes.get("tool.name")):
+        if tool_name := _real_tool_name(span):
             try:
-                input_value: Any = span.attributes.get("input.value", {})
+                input_value: Any = span.attributes.get("input.value", {})  # type: ignore[union-attr]
```
ajay-kesavan: remove the comment

packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
```diff
@@ -87,8 +101,8 @@ def extract_tool_calls_outputs(spans: Sequence[ReadableSpan]) -> list[ToolOutput
     potential_output_keys = ["content"]
     tool_calls_outputs = []
     for span in spans:
-        if span.attributes and (tool_name := span.attributes.get("tool.name")):
-            output = span.attributes.get("output.value", "")
+        if tool_name := _real_tool_name(span):
+            output = span.attributes.get("output.value", "")  # type: ignore[union-attr]
```
ajay-kesavan: remove comment

packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
ajay-kesavan: 🤖 [Claude] Done in dee54934 — docstring is one line now.

packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
ajay-kesavan: 🤖 [Claude] Done in dee54934 — predicate now returns the attrs Mapping, so each extractor reads from a locally-narrowed attrs var and the # type: ignore[union-attr] is no longer needed.

packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
ajay-kesavan: 🤖 [Claude] Done in dee54934 — same refactor as the line-74 thread; output = attrs.get(...) now reads from the narrowed Mapping so the # type: ignore drops.

ameyjain (approved):

Chibionos (approved):

packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
```diff
@@ -55,20 +62,19 @@ def extract_tool_calls(spans: Sequence[ReadableSpan]) -> list[ToolCall]:
...
-        if span.attributes and (tool_name := span.attributes.get("tool.name")):
+        if (attrs := _real_tool_attrs(span)) is not None:
+            tool_name = str(attrs["tool.name"])
             try:
-                input_value: Any = span.attributes.get("input.value", {})
-                # Ensure input_value is a string before parsing
+                input_value: Any = attrs.get("input.value", {})
                 if isinstance(input_value, str):
                     arguments = ast.literal_eval(input_value)
                 elif isinstance(input_value, dict):
```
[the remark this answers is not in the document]
ajay-kesavan: 🤖 [Claude] Addressed in PR #1725 (commit c646bc13) — `extract_tool_calls` now tries `json.loads` first (handles `true`/`false`/`null` and double-quoted keys), falls back to `ast.literal_eval` for the historical single-quoted Python dict repr, and rejects non-dict parse results to keep `ToolCall.args` type-correct. The outer `except` is now `(SyntaxError, ValueError)` — `ValueError` covers `json.JSONDecodeError` (it's a subclass) and `SyntaxError` covers `ast.literal_eval` on garbage input. Three new tests in PR #1725 cover the JSON-literals path, the Python-literal fallback, and the non-dict guard.

Landed it on PR #1725 since that PR already modifies `extract_tool_calls` (the `include_args` parameter); avoids re-invalidating @ameyjain's approval on this PR.

PR #1728, opened 2026-06-17
feat(agent): add ontology binding to Data Fabric context config

sankalp-uipath:
## What
Adds Data Fabric ontology binding to the agent context model: a context can carry one or more ontologies that the runtime fetches to ground Text2SQL.

- `agent/models/agent.py` — `AgentContextResourceConfig.ontology_set` (alias `ontologySet`), a list of `DataFabricOntologyItem` nested in the context alongside `entitySet`.
- `platform/entities/_entity_ontology_service.py` — `DataFabricOntologyItem` (`{name, folderId, description?}`) + `EntityOntologyService`, which fetches one ontology file: `GET datafabric_/api/ontologies/{name}/files/{type}` on the authenticated, folder-scoped client.
- `platform/entities/_entities_service.py` — `get_ontology_file_async(ontology_name, file_type="owl", folder_key=None)` is a thin public wrapper delegating to `EntityOntologyService` (all ontology logic lives in the ontology sub-service). One file per call; the langchain consumer fetches `owl` and `r2rml` as separate calls.

## Why
An ontology is part of a Data Fabric context — domain knowledge over the same entities — so it is configured inline on the context (a sibling of `entitySet`), mirroring how entities are bound. Each ontology carries its own `folderId` so it can live in a different folder from the entities.

## Notes
- **Additive + backward compatible:** `ontology_set` defaults to `None`.
- **No `referenceKey` on ontology items:** ontologies are not solution-aware at this point, so there is no solution key to reference. The field is dropped from `DataFabricOntologyItem` — it was never populated or read (ontologies resolve by `name` + `folderId`). Backward compatible: configs that still send `referenceKey` are tolerated via `extra="allow"`.
- **Validation is server-side:** file-type/name validation lives in the ontology API (QE), not the client — the SDK forwards the request and surfaces the API's rejection (e.g. an unsupported file type) as `EnrichedException` (status code + response body + extracted message) rather than swallowing it. Redundant client-side allowlists were removed per review.
- **Tests:** nested `ontologySet` parse/dump, required `folderId` per item, backward-compat default, full-`AgentDefinition` normalization; `get_ontology_file_async` covered for spec, folder header, allowed file types, and the negative case (API rejection of an unsupported file type surfaces as `EnrichedException`).
- Consumed by uipath-langchain PR #911.

## Release
Shipped as **`uipath` 2.13.2** and **`uipath-platform` 0.2.1** (published to PyPI). Consumers should pin **`uipath-platform>=0.2.1`** (and `uipath>=2.13.2`) to get `ontologySet` / `AgentContextType.DATA_FABRIC_ONTOLOGY` / `get_ontology_file_async`.


packages/uipath/src/uipath/agent/models/agent.py
```diff
@@ -440,6 +443,11 @@ class AgentContextResourceConfig(BaseAgentResourceConfig):
         None, description="Context settings"
     )
     entity_set: Optional[List[DataFabricEntityItem]] = Field(None, alias="entitySet")
+    ontology: Optional[DataFabricOntologyItem] = Field(
```
milind-jain-uipath: List

milind-jain-uipath (changes requested):

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -58,6 +61,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Ontology name contract (QueryEngine OntologyController): lowercase, starts
+# with a letter, max 64 chars. The name becomes a URL path segment.
+_ONTOLOGY_NAME_RE = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
+# Allowed ontology component file types (also URL path segments).
+_ONTOLOGY_FILE_TYPES = frozenset({"owl", "r2rml", "shacl", "summary", "context"})
```
UIPath-Harshit: Whats the rationale here?

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -1100,6 +1109,68 @@ async def delete_record_async(self, entity_key: str, record_id: str) -> None:
...
+        file_type: str = "owl",
+        folder_key: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """Fetch one file of an ontology from Data Fabric.
+
+        !!! warning "Preview Feature"
+            This method is currently experimental. Behavior and parameters are
+            subject to change in future versions.
+
+        Ontologies are served by the same QueryEngine service as entity SQL
```
UIPath-Harshit: Remove this implementation detail from the doc

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -1100,6 +1109,68 @@ async def delete_record_async(self, entity_key: str, record_id: str) -> None:
...
+        Returns:
+            Dict[str, Any]: The file record (e.g. ``content``, ``mediaType``).
+
+        Raises:
+            ValueError: If the ontology name or file type is invalid.
+        """
+        self._validate_ontology_name(ontology_name)
+        self._validate_file_type(file_type)
+        spec = self._ontology_file_spec(ontology_name, file_type)
+        headers = {"Accept": "application/json", **header_folder(folder_key, None)}
```
UIPath-Harshit: Do you need this? Generally its handled by the spec method through base class.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
```diff
@@ -1100,6 +1109,68 @@ async def delete_record_async(self, entity_key: str, record_id: str) -> None:
...
+        response = await self.request_async(spec.method, spec.endpoint, headers=headers)
+        return response.json()
+
+    @staticmethod
+    def _validate_ontology_name(ontology_name: str) -> None:
```
UIPath-Harshit: All these helper methods needs to move away from this file. I would suggest to create entities_ontology_service separately and consume just the ontology methods from this class.

packages/uipath-platform/src/uipath/platform/entities/entities.py
```diff
@@ -843,6 +843,26 @@ class DataFabricEntityItem(BaseModel):
     description: Optional[str] = None
 
 
+class DataFabricOntologyItem(BaseModel):
```
UIPath-Harshit: Same for this. Move it to entities_ontology_service

packages/uipath-platform/src/uipath/platform/entities/entities.py
```diff
@@ -843,6 +843,26 @@ class DataFabricEntityItem(BaseModel):
...
+    ``folderId`` so it resolves from its own folder. ``name`` is used to fetch
+    the ontology from the QueryEngine ontology API.
+    """
+
+    model_config = ConfigDict(
+        validate_by_name=True, validate_by_alias=True, extra="allow"
+    )
+
+    name: str
+    ontology_key: Optional[str] = Field(None, alias="referenceKey")
```
UIPath-Harshit: Ontology is not solution aware at this point. Does it make sense for this to come from here?

packages/uipath/src/uipath/agent/models/agent.py
```diff
@@ -440,6 +440,15 @@ class AgentContextResourceConfig(BaseAgentResourceConfig):
         None, description="Context settings"
     )
     entity_set: Optional[List[DataFabricEntityItem]] = Field(None, alias="entitySet")
+    ontology_set: Optional[List[DataFabricOntologyItem]] = Field(
```
UIPath-Harshit, line 444: Is the plan to add set of ontologies or single ontology ? Set doesnt make sense to me , but let me know if that decision has already been settled. cc: @milind-jain-uipath

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
sankalp-uipath: removed them , my bad, QE already does these validations , so this was redundant.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
sankalp-uipath: fixed.

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
sankalp-uipath: Agreed, removed the redundant accept, and moved folder headers to request spec. please review again.

packages/uipath-platform/src/uipath/platform/entities/entities.py
sankalp-uipath: fixed

packages/uipath-platform/src/uipath/platform/entities/_entities_service.py
sankalp-uipath: agreed
Among the helpers method validate name and filetype are deleted as per the previous comment, I have moved spec method to _entity_ontology_service and also added fetch logic there
Now get_ontology_file_async is a just thin method calling method from _entity_ontology_service.

packages/uipath/src/uipath/agent/models/agent.py
sankalp-uipath: Please correct me if am wrong @milind-jain-uipath 
Yes plan is to allow users to add set of ontologies

packages/uipath-platform/src/uipath/platform/entities/entities.py
sankalp-uipath: Sorry I didn't get you why ontology not being solution aware will cause issue, currently folder id is just for scoping and coming from agent.json

packages/uipath-platform/tests/services/test_entities_service.py
```diff
@@ -2647,3 +2647,60 @@ def test_5xx_with_batch_shape_still_propagates(
...
+
+        headers = service._ontology.request_async.call_args.kwargs["headers"]
+        assert "x-uipath-folderkey" not in headers
+
+    @pytest.mark.anyio
+    @pytest.mark.parametrize(
+        "file_type", ["owl", "r2rml", "shacl", "summary", "context"]
+    )
+    async def test_accepts_allowed_file_types(
+        self, service: EntitiesService, file_type: str
```
andreitava-uip, line 2711: nit: how about a test for the negative case:
Rejects unsupported file types?

packages/uipath-platform/src/uipath/platform/entities/entities.py
sankalp-uipath: Removed ontology_key, please review again.

packages/uipath-platform/src/uipath/platform/entities/_entity_ontology_service.py
```diff
@@ -0,0 +1,80 @@
...
+        self,
+        config: UiPathApiConfig,
+        execution_context: UiPathExecutionContext,
+        folders_service: Optional[FolderService] = None,
+    ) -> None:
+        """Initialise the ontology service."""
+        super().__init__(config=config, execution_context=execution_context)
+        self._folders_service = folders_service
+
+    async def get_file_async(
```
UIPath-Harshit, line 57: Should be called get_onotology_bundle_async

milind-jain-uipath (approved):

PR #1731, opened 2026-06-17
feat(platform): include jobKey in LLM Gateway trace baggage

RunnanJia:
## Why
`build_trace_context_headers` (`uipath/platform/chat/llm_trace_context.py`) propagates `folderKey`, `agentId`, and `processKey` into `x-uipath-tracebaggage`, but **not `jobKey`** — even though `UiPathConfig.job_key` is available. LLM Gateway now builds the model-call span from this baggage (see the LLMGateway span work), so a missing `jobKey` means the span can't be attributed to the originating job.

Today callers compensate by hand-injecting `jobKey` through `extra_baggage` (e.g. uipath-agents-python). Emitting it natively here fixes it once for **every** caller and lets those workarounds be removed.

## What
One addition in `build_trace_context_headers`, mirroring the existing `processKey` block:
```python
if job_key := UiPathConfig.job_key:
    baggage_parts.append(f"jobKey={job_key}")
```
Gated by the existing `EnableTraceContextHeaders` flag like the rest of the function; no behavior change when the flag is off.

## Note
Callers that already inject `jobKey` via `extra_baggage` will momentarily send it twice in the comma-joined header; the LLM Gateway baggage parser is last-write-wins, so this is harmless. Those callers can drop the manual injection once this ships.

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
+++ b/packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
@@ -34,8 +34,14 @@ def build_trace_context_headers(
     ctx = span.get_span_context()
     if config_trace_id and ctx and ctx.span_id:
         trace_id = _SpanUtils.normalize_trace_id(config_trace_id)
-        span_id = format(ctx.span_id, "032x")
-        headers["x-uipath-traceparent-id"] = f"00-{trace_id}-{span_id}"
+        # An OTEL span id is 64-bit => 16 lowercase hex chars, and the W3C traceparent
+        # requires the trailing trace-flags segment. The LLM Gateway's strict parser
+        # (UiPath.Tracing.TraceParent.TryParse) rejects the header unless it is exactly
+        # {version}-{32-hex trace}-{16-hex span}-{2-hex flags}; on rejection the gateway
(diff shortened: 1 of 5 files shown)


ctiliescuuipath (approved):

packages/uipath-platform/src/uipath/platform/chat/llm_trace_context.py
```diff
@@ -44,6 +44,8 @@ def build_trace_context_headers(
         baggage_parts.append(f"agentId={agent_id}")
     if process_uuid := UiPathConfig.process_uuid:
         baggage_parts.append(f"processKey={process_uuid}")
+    if job_key := UiPathConfig.job_key:
+        baggage_parts.append(f"jobKey={job_key}")
```
ctiliescuuipath, line 54: do we need this as part of baggage. I saw that on other flows this is pass as a independent header https://github.com/UiPath/uipath-python/pull/1739 (how is also licensingRefId)

RunnanJia: Yes, as i mentioned in another PR, we want all these info which will write into attributes in the baggage, but still if we saw conflicts with other headers, the stand alone headers are the source of truth.

PR #1738, opened 2026-06-22
feat(platform): add GovernanceService for agenticgovernance_ ingress

viswa-uipath:
## Summary

Platform-side foundation for UiPath agent governance — protocols in `uipath-core` and a concrete `GovernanceService` + provider adapter in `uipath-platform`. The runtime-side wrapper that consumes these will land in `uipath-runtime` in a follow-up PR.

- **`uipath-core`** — declares `GovernancePolicyProvider` / `GovernanceCompensationProvider` protocols (`runtime_checkable`) alongside `EvaluatorProtocol`. Owns the wire-format models `PolicyContext`, `PolicyResponse`, `FiredRule`, `GovernRequest`. No registry or concrete provider — `uipath-core` does not depend on `uipath-platform` or `uipath-runtime`.
- **`uipath-platform`** — adds `GovernanceService` (`sdk.governance`) for the org-scoped `agenticgovernance_` ingress with public ergonomic methods:
  - `retrieve_policy(*, is_conversational=None)` + async — GET the tenant policy pack.
  - `compensate(*, hook, validators, rules, data, trace_id, src_timestamp, agent_name, runtime_id, folder_key=None, job_key=None, process_key=None, reference_id=None, agent_version=None)` + async — POST the compensating `/runtime/govern` call; builds `GovernRequest` internally; job-context fields auto-fill from `UiPathConfig` with caller values (including empty strings) taking precedence over the env-backed fallback.
  - Honors `UIPATH_SERVICE_URL_AGENTICGOVERNANCE` for local dev — the override path injects routing headers since the platform router is bypassed.
- **`uipath-platform`** — adds `UiPathPlatformGovernanceProvider` as the dedicated adapter that satisfies both core protocols by delegating to the service. Runtime consumers receive this object via constructor injection without importing `uipath-platform` directly. Construct via `UiPathPlatformGovernanceProvider(service=UiPath().governance)` or by passing `config` / `execution_context`.
- **`uipath-platform`** — adds public `resolve_trace_id(fallback)` helper in `common/_base_service.py` so callers can capture the trace id on the hook thread before hopping to a background pool.

Bumps: `uipath-core` `0.5.20 → 0.5.21`, `uipath-platform` `0.1.71 → 0.1.73`, `uipath` `2.11.6 → 2.11.9`.

## Motivation

Replaces hand-rolled `urllib` HTTP in [`uipath-runtime-python` PRs #121](https://github.com/UiPath/uipath-runtime-python/pull/121) and [#123](https://github.com/UiPath/uipath-runtime-python/pull/123) per [radu-mocanu's review comment](https://github.com/UiPath/uipath-runtime-python/pull/121#discussion_r3448412376): *"tenant/org/access token etc. are platform concerns. The runtime's concern is just managing the policies."*

The runtime can't import `uipath-platform` (dependency direction is the other way), so the protocols live in `uipath-core`. The runtime layer will accept the provider via constructor injection — no global registry, no hidden imports across the layer boundary.

After this lands (and the follow-up runtime PR), the runtime gives up `urllib`, the browser-UA WAF workaround, manual `UIPATH_ACCESS_TOKEN` reads, manual org/tenant resolution, and the import-time `uipath-platform` dependency — `BaseService` provides all of that, and the runtime depends only on `uipath-core` protocols + a constructor-injected provider.

## Design choices

- **Protocols in `uipath-core`, concrete provider in `uipath-platform`** — keeps `uipath-core` as the lowest layer with no upward dependencies; runtime consumers depend only on the protocols. No module-level provider registry — providers move via constructor injection.
- **`runtime_checkable` Protocols** — any object exposing `get_policy(context)` / `compensate(request)` satisfies the contract.
- **Public SDK takes individual params; provider takes the object** — `compensate(*, hook, ...)` is the ergonomic SDK shape that matches `retrieve_policy`'s pattern and stays consistent with the rest of the platform services. The internal `_compensate(request)` is the worker the provider adapter calls to satisfy the protocol shape without unpacking.
- **`UiPathPlatformGovernanceProvider` is the dedicated adapter** — implements both core protocols by delegating to `GovernanceService`. The service is the SDK surface; the provider is the runtime-facing contract.
- **URL composed directly** (not via `RequestSpec` + `Endpoint`): the `agenticgovernance_` ingress uses `{origin}/{org_id_uuid}/agenticgovernance_/...` — the org UUID, not the slug `UiPathUrl.scope_url` produces. Documented in `_build_org_scoped_request`.
- **Honors `UIPATH_SERVICE_URL_AGENTICGOVERNANCE` for local dev** — `_build_org_scoped_request` resolves the override itself (mirroring what `BaseService.request` does for paths that fit `scope_url`) and injects routing headers, because the org-UUID URL shape can't go through the standard `scope_url` flow.
- **`is_conversational: bool | None`** rather than an enum: matches the runtime's existing `_agent_is_conversational` flag. The `True/False/None → conversational/autonomous/omit` wire mapping is an implementation detail of the service.
- **`resolve_trace_id` returns the value**, doesn't mutate headers: the existing `_inject_trace_context` only injects the header. Compensation needs the id in the request body, and must resolve on the hook thread before the pool hop because OpenTelemetry context is thread-local.
- **`@cached_property` on `sdk.governance`**: matches `buckets`/`attachments`/`folders`. Avoids fresh `httpx.Client`/`AsyncClient` per access in the background compensation loop.
- **`PolicyResponse.mode: EnforcementMode | None` with lenient coercion**: a `@field_validator(mode="before")` maps unknown wire values to `None` instead of raising, so a server-side mode addition can't break agent startup.

## Out of scope (follow-up PRs)

- **Runtime wrapper**: `GovernanceRuntime` (calls `provider.get_policy()` at execute/stream start, exposes `compensate()` to agent code) will land in `uipath-runtime`, not here. CLI wiring (`cli_run` / `cli_debug` constructing the runtime with the injected provider) lives there too.
- **Removing the `EnablePythonGovernanceChecker` feature flag**: kept in place until test-pypi rollout finishes; will be removed in a separate PR.

## Test plan

- [x] `ruff check .` / `ruff format --check .` — clean across all three packages
- [x] `mypy src tests` — clean
- [x] `pytest --no-cov` — full suite green
- [x] **31** uipath-core tests (`tests/governance/test_providers.py`): protocol conformance (`runtime_checkable` accept/reject), wire-format models (defaults, camelCase aliasing + intentional snake_case `src_timestamp`, optional job-context excluded when `None`, lenient `EnforcementMode` coercion).
- [x] **37** uipath-platform tests across `tests/services/test_governance_service.py` (28) and `tests/services/test_governance_provider.py` (9): policy fetch happy path / defaults / header + bearer wiring / `agentType` query param (parameterized True/False) / missing org-tenant / HTTP errors / async; compensate payload aliasing / job-context auto-fill from env / caller override precedence / caller empty-string preserved over env / omitted keys / HTTP errors / async; `GovernanceService.get_policy` satisfies the policy protocol; `UiPathPlatformGovernanceProvider` construction (existing service, built-from-config, requires service or full kwargs) + both-protocol conformance + delegation; `UIPATH_SERVICE_URL_AGENTICGOVERNANCE` redirects both policy fetch and compensate POST; 5 `resolve_trace_id` cases.
- [ ] Integration: runtime-side PRs (`uipath-runtime-python` #121/#123) to be refactored to consume the constructor-injected provider — separate PRs.

## Development Packages

> Dev-version numbers below are tied to a previous CI build; the next CI run on this branch will regenerate them.

### uipath-core

```toml
[project]
dependencies = [
  "uipath-core==0.5.21.dev1017386896",

  # "uipath-core>=0.5.21.dev1017380000,<0.5.21.dev1017390000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
uipath-core = { index = "testpypi" }
```

### uipath

```toml
[project]
dependencies = [
  "uipath==2.11.9.dev1017386896",

  # "uipath>=2.11.9.dev1017380000,<2.11.9.dev1017390000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
uipath-core = { index = "testpypi" }

[tool.uv]
override-dependencies = ["uipath-platform==0.1.73.dev1017386896", "uipath-core==0.5.21.dev1017386896"]
```

### uipath-platform

```toml
[project]
dependencies = [
  "uipath-platform==0.1.73.dev1017386896",

  # "uipath-platform>=0.1.73.dev1017380000,<0.1.73.dev1017390000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
uipath-core = { index = "testpypi" }

[tool.uv]
override-dependencies = ["uipath-core==0.5.21.dev1017386896"]
```

### DIFF
--- a/packages/uipath-core/src/uipath/core/governance/providers.py
+++ b/packages/uipath-core/src/uipath/core/governance/providers.py
@@ -1,0 +1,153 @@
+"""Provider protocols for governance backend interactions.
+
+The runtime needs two backend interactions to function:
+
+- Fetching the policy pack at startup.
+- Firing the compensating ``/runtime/govern`` POST when a
+  ``guardrail_fallback`` rule matches so the server can run the disabled
+  centralised guardrail and write the per-rule LLMOps audit records.
+
+Both have wire formats owned by the ``agenticgovernance_`` ingress.
+Defining the contracts here — alongside :class:`EvaluatorProtocol` —
+lets runtime consumers depend on stable protocols and receive a
+concrete provider via constructor injection. Concrete providers live
+outside this package; ``uipath-core`` does not import them.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Protocol, runtime_checkable
+
+from pydantic import BaseModel, ConfigDict, Field, field_validator
+
+from .models import EnforcementMode
+
+# ----------------------------------------------------------------------
+# Wire-format models
+# ----------------------------------------------------------------------
+
+
+class PolicyContext(BaseModel):
+    """Caller-supplied selectors for the policy fetch.
+
+    Wrapping the selectors in a model keeps the protocol surface stable
+    when the server grows new selector dimensions — adding a field here
+    doesn't change :meth:`GovernancePolicyProvider.get_policy`.
+
+    Today carries only :attr:`is_conversational`; future selectors land
+    here.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    is_conversational: bool | None = None
+
+
+class PolicyResponse(BaseModel):
+    """Parsed governance backend response.
+
+    Wire envelope::
+
+        {
+            "mode": "audit" | "enforce" | "disabled",
+            "policies": "<YAML string>"
+        }
+
+    Attributes:
+        mode: Platform-controlled enforcement mode for the tenant. May
+            be ``None`` when the backend omits it. A wire value the SDK
+            doesn't know about parses as ``None`` rather than raising,
+            so a server-side mode addition can't break agent startup.
+        policies: Policy pack YAML the caller compiles into its policy
+            index. May be an empty string when no rules are configured.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    mode: EnforcementMode | None = Field(default=None)
+    policies: str = Field(default="")
+
+    @field_validator("mode", mode="before")
+    @classmethod
+    def _coerce_mode(cls, value: object) -> EnforcementMode | None:
+        if value is None or isinstance(value, EnforcementMode):
+            return value
+        try:
+            return EnforcementMode(value)
+        except ValueError:
+            return None
+
+
+class FiredRule(BaseModel):
+    """Per-rule metadata carried in the ``/runtime/govern`` payload.
+
+    One entry per matching ``guardrail_fallback`` condition. The server
+    writes one LLMOps trace record per entry, so callers must include
+    every fired rule even when multiple share the same ``validator``.
+    """
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    rule_id: str = Field(alias="ruleId")
+    rule_name: str = Field(alias="ruleName")
+    pack_name: str = Field(alias="packName")
+    validator: str
+
+
+class GovernRequest(BaseModel):
+    """Request body for the ``/runtime/govern`` compensating governance POST.
+
+    Field aliases match the on-the-wire JSON keys. ``src_timestamp`` is
+    snake_case on the wire (intentional — preserved verbatim); every
+    other key is camelCase.
+
+    Job-context fields (``folder_key`` / ``job_key`` / ``process_key`` /
+    ``reference_id`` / ``agent_version``) are optional; callers omit
+    them by leaving them ``None``. How unset fields are resolved (e.g.
+    auto-filled from environment) is the concrete provider's concern,
+    not part of this wire contract.
+    """
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    validators: list[str] = Field(alias="type")
+    rules: list[FiredRule]
+    data: dict[str, Any]
+    hook: str
+    trace_id: str = Field(alias="traceId")
+    src_timestamp: str  # wire key is intentionally snake_case
+    agent_name: str = Field(alias="agentName")
+    runtime_id: str = Field(alias="runtimeId")
+
+    folder_key: str | None = Field(default=None, alias="folderKey")
+    job_key: str | None = Field(default=None, alias="jobKey")
+    process_key: str | None = Field(default=None, alias="processKey")
+    reference_id: str | None = Field(default=None, alias="referenceId")
+    agent_version: str | None = Field(default=None, alias="agentVersion")
+
+
+# ----------------------------------------------------------------------
+# Provider protocols
+# ----------------------------------------------------------------------
+
+
+@runtime_checkable
+class GovernancePolicyProvider(Protocol):
+    """Contract for fetching the governance policy pack.
+
+    Any object exposing a ``get_policy(context) -> PolicyResponse``
+    method satisfies this protocol.
+    """
+
+    def get_policy(self, context: PolicyContext) -> PolicyResponse:
+        """Fetch the policy pack for the active org/tenant."""
+        ...
+
+
+@runtime_checkable
+class GovernanceCompensationProvider(Protocol):
+    """Contract for firing the compensating ``/runtime/govern`` POST."""
+
+    def compensate(self, request: GovernRequest) -> None:
+        """Fire the compensating governance POST. Fire-and-forget."""
+        ...
--- a/packages/uipath-platform/src/uipath/platform/_uipath.py
+++ b/packages/uipath-platform/src/uipath/platform/_uipath.py
@@ -22,6 +22,7 @@ from .context_grounding import ContextGroundingService
 from .documents import DocumentsService
 from .entities import EntitiesService
 from .errors import BaseUrlMissingError, SecretMissingError
+from .governance import GovernanceService
 from .guardrails import GuardrailsService
 from .memory import MemoryService
 from .orchestrator import (
@@ -168,6 +169,10 @@ class UiPath:
     @property
     def guardrails(self) -> GuardrailsService:
         return GuardrailsService(self._config, self._execution_context)
+
+    @cached_property
+    def governance(self) -> GovernanceService:
+        return GovernanceService(self._config, self._execution_context)
 
     @property
     def agenthub(self) -> AgentHubService:
--- a/packages/uipath-platform/src/uipath/platform/governance/_governance_service.py
+++ b/packages/uipath-platform/src/uipath/platform/governance/_governance_service.py
@@ -1,0 +1,356 @@
+"""Service for the ``agenticgovernance_`` ingress.
+
+Wraps the two governance backend endpoints UiPath exposes:
+
+- ``GET  /{org}/agenticgovernance_/api/v1/runtime/policy``  — fetch the
+  tenant-managed policy pack (see :meth:`GovernanceService.retrieve_policy`).
+- ``POST /{org}/agenticgovernance_/api/v1/runtime/govern``  — compensating
+  governance call fired when a ``guardrail_fallback`` rule matches
+  (see :meth:`GovernanceService.compensate`).
+
+Org/tenant scoping is read from :class:`UiPathConfig`; auth, retries,
+trace context, and error enrichment come from :class:`BaseService`.
+"""
+
+from typing import Any, Optional
+
+from uipath.core import traced
+from uipath.core.governance import (
+    FiredRule,
+    GovernRequest,
+    PolicyContext,
+    PolicyResponse,
+)
+
+from ..common._base_service import BaseService
+from ..common._config import UiPathConfig
+from ..common._service_url_overrides import (
+    inject_routing_headers,
+    resolve_service_url,
+)
+from ..common.constants import HEADER_INTERNAL_TENANT_ID
+
+# The agenticgovernance_ ingress lives at a separate org-scoped path that
+# uses the organization UUID (not the slug exposed by ``UIPATH_URL``).
+GOVERNANCE_SERVICE_PREFIX = "agenticgovernance_"
+POLICY_API_PATH = "api/v1/runtime/policy"
+GOVERN_API_PATH = "api/v1/runtime/govern"
+AGENT_TYPE_PARAM = "agentType"
+
+
+class GovernanceService(BaseService):
+    """Service for the agenticgovernance_ ingress.
+
+    Exposes two endpoints:
+
+    - :meth:`retrieve_policy` — GET the tenant-managed policy pack.
+    - :meth:`compensate` — POST a compensating ``/runtime/govern`` call
+      so the server can run a disabled centralized guardrail and write
+      the per-rule LLMOps audit records itself.
+
+    Org and tenant scoping come from :attr:`UiPathConfig.organization_id`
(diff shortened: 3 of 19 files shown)


packages/uipath-core/src/uipath/core/governance/providers.py
```diff
@@ -0,0 +1,158 @@
...
+
+- Fetching the policy pack at startup.
+- Firing the compensating ``/runtime/govern`` POST when a
+  ``guardrail_fallback`` rule matches so the server can run the disabled
+  centralised guardrail and write the per-rule LLMOps audit records.
+
+Both have wire formats owned by the platform's ``agenticgovernance_``
+ingress. Defining the contracts here in ``uipath-core`` — alongside
+:class:`EvaluatorProtocol` — lets the runtime depend only on the lowest
+layer; the HTTP implementation lives in ``uipath-platform`` (or any
```
radu-mocanu: please remove all references to `uipath-platform/` `uipath-runtime` from core library. `uipath-core` doesn t need to know about these
```suggestion
layer;` (or any
```

packages/uipath-core/src/uipath/core/governance/providers.py
```diff
@@ -0,0 +1,158 @@
...
+other implementation a deployment chooses) and is supplied to runtime
+consumers via constructor injection.
+
+Concrete platform-backed implementations:
+:class:`uipath.platform.governance.UiPathPlatformGovernanceProvider`
```
radu-mocanu: same as above. we can cross-reference just the other way around (i.e. from runtime to core)

packages/uipath-core/src/uipath/core/governance/providers.py
```diff
@@ -0,0 +1,158 @@
...
+    when the server grows new selector dimensions — adding a field here
+    doesn't change :meth:`GovernancePolicyProvider.get_policy`.
+
+    Today carries only :attr:`is_conversational`; future selectors land
+    here.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    is_conversational: Optional[bool] = None
```
radu-mocanu: nit (we prefer the new representation mode)
```suggestion
    is_conversational: bool | None = None
```

packages/uipath-platform/src/uipath/platform/governance/_governance_service.py
```diff
@@ -0,0 +1,250 @@
...
+                "to be set in the environment."
+            )
+        tenant_id = UiPathConfig.tenant_id
+        if not tenant_id:
+            raise ValueError(
+                "Governance call requires UIPATH_TENANT_ID "
+                "to be set in the environment."
+            )
+
+        url = (
```
radu-mocanu, line 314: does this still work with service url overrides? `_build_org_scoped_request()` returns an absolute URL, but `resolve_service_url()` expects a relative service-prefixed path like `/agenthub_/...`, so `UIPATH_SERVICE_URL_AGENTICGOVERNANCE` will not be applied.

packages/uipath-core/src/uipath/core/governance/providers.py
```diff
@@ -0,0 +1,152 @@
...
+
+@runtime_checkable
+class GovernancePolicyProvider(Protocol):
+    """Contract for fetching the governance policy pack.
+
+    Any object exposing a ``get_policy(context) -> PolicyResponse``
+    method satisfies this protocol.
+    """
+
+    def get_policy(self, context: PolicyContext) -> PolicyResponse:
```
radu-mocanu, line 142: question on the shape here. if we fetch policies only once when agent execution starts, I would rather do that in the CLI/composition layer and pass the resolved `PolicyResponse` through the runtime ctor. if the runtime needs to refresh or fetch differently later, then the provider protocol makes sense. which one is the intended lifecycle?

packages/uipath/src/uipath/_governance/setup.py
```diff
@@ -0,0 +1,67 @@
...
+    Args:
+        delegate: The runtime to wrap.
+        policy_context: Selectors forwarded to the policy fetch. Defaults
+            to :class:`PolicyContext()` (server applies its defaults).
+
+    Returns:
+        Either a :class:`GovernanceRuntime` wrapping ``delegate``, or
+        ``delegate`` itself when governance can't be wired.
+    """
+    if not os.environ.get("UIPATH_URL") or not os.environ.get("UIPATH_ACCESS_TOKEN"):
```
radu-mocanu: should this also check `is_governance_enabled()`? core defines `EnablePythonGovernanceChecker` as off by default, but here any run with `UIPATH_URL` + `UIPATH_ACCESS_TOKEN` will wrap and call `/runtime/policy`.

packages/uipath-platform/src/uipath/platform/governance/_governance_service.py
```diff
@@ -0,0 +1,250 @@
...
+    """Service for the agenticgovernance_ ingress.
+
+    Exposes two endpoints:
+
+    - :meth:`retrieve_policy` — GET the tenant-managed policy pack.
+    - :meth:`compensate` — POST a compensating ``/runtime/govern`` call
+      so the server can run a disabled centralized guardrail and write
+      the per-rule LLMOps audit records itself.
+
+    Also satisfies the
```
radu-mocanu: nit: I would keep this service docstring about the platform service only. details like the CLI registering it or runtime consuming it through protocols are composition-layer details and can get stale here. this service should just document the platform operations it exposes and the auth/config behavior it owns.

packages/uipath-platform/src/uipath/platform/governance/_governance_service.py
```diff
@@ -0,0 +1,250 @@
...
+    async def get_policy_async(self, context: PolicyContext) -> PolicyResponse:
+        """Async variant of :meth:`get_policy`."""
+        return await self.retrieve_policy_async(
+            is_conversational=context.is_conversational
+        )
+
+    # ── Compensating governance call ─────────────────────────────────
+
+    @traced(name="governance_compensate", run_type="uipath")
+    def compensate(self, request: GovernRequest) -> None:
```
radu-mocanu: ]he service API should not expose this as a `GovernRequest` object?.for the provider protocol this makes sense, but `GovernanceService` is the platform SDK surface. to be consistent with other services, I would expect this method to take the meaningful params and build the request/payload internally, or keep the object-taking method internal/provider-only.

packages/uipath-core/src/uipath/core/governance/providers.py
viswa-uipath: Since we are in AUDIT mode and we have only one pack, we are making the call at first. Going forward we might change this approach to multi call based on the enforcement logic. Lets keep it as a protocol for extensibility.

packages/uipath-platform/src/uipath/platform/_uipath.py
```diff
@@ -169,6 +170,20 @@ def mcp(self) -> McpService:
     def guardrails(self) -> GuardrailsService:
         return GuardrailsService(self._config, self._execution_context)
 
+    @cached_property
+    def governance(self) -> GovernanceService:
+        return GovernanceService(self._config, self._execution_context)
+
+    @cached_property
+    def governance_provider(self) -> UiPathPlatformGovernanceProvider:
```
radu-mocanu: no reason to expose this via UiPath object. just the service is fine

packages/uipath/src/uipath/_governance/setup.py
viswa-uipath: We will remove the FF altogether as it will not work for coded agents without user defined flags. This is currently in place till we do test pypi.

packages/uipath/src/uipath/_governance/setup.py
```diff
@@ -0,0 +1,67 @@
...
+        logger.debug(
+            "Skipping governance wrapping: UIPATH_URL / UIPATH_ACCESS_TOKEN not set."
+        )
+        return delegate
```
radu-mocanu: it is not CLI's concern to check this. if the env vars are not set the execution will likely fail anyways. please remove this

packages/uipath/src/uipath/_governance/runtime.py
```diff
@@ -0,0 +1,190 @@
...
+    UiPathRuntimeResult,
+    UiPathStreamOptions,
+)
+from uipath.runtime.events import UiPathRuntimeEvent
+from uipath.runtime.schema import UiPathRuntimeSchema
+
+logger = logging.getLogger(__name__)
+
+
+class GovernanceRuntime:
```
radu-mocanu: the runtime should live in uipath-runtime. in the cli_run/cli_debug we just construct that object by giving it the providers and policy_contexts. this class should be moved to `uipath-runtime`

packages/uipath/src/uipath/_governance/setup.py
radu-mocanu: the feature should be opt-in for coded agents. default false -> if user sets the experimental env var then the runtime should be applied

packages/uipath-platform/src/uipath/platform/governance/_governance_service.py
radu-mocanu: check how other services APIs look like

packages/uipath/src/uipath/_governance/setup.py
viswa-uipath: We will control this via backend feature flag for few organizations (For private preview). This is not an optional feature. Governance is a platform concern. Users cant opt out if the admin enables it.

packages/uipath-core/src/uipath/core/governance/providers.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class GovernRequest(BaseModel):
+    """Request body for the ``/runtime/govern`` compensating governance POST.
+
+    Field aliases match the on-the-wire JSON keys. ``src_timestamp`` is
+    snake_case on the wire (intentional — preserved verbatim); every
+    other key is camelCase.
+
+    Job-context fields (``folder_key`` / ``job_key`` / ``process_key`` /
```
radu-mocanu, line 104: nit: remove the `UiPathConfig` reference from core docs. this model should describe only the optional request contract; the fact that platform fills missing values from `UiPathConfig` belongs in the platform service docs.

radu-mocanu (approved):

packages/uipath-platform/tests/services/test_governance_provider.py
```diff
@@ -0,0 +1,166 @@
...
+)
+
+from uipath.platform import UiPathApiConfig, UiPathExecutionContext
+from uipath.platform.governance import (
+    GovernanceService,
+    UiPathPlatformGovernanceProvider,
+)
+
+ORG_ID = "62d19041-d1aa-454d-958d-1375329845dc"
+TENANT_ID = "e150b32b-8815-4560-8243-055ffc9b7523"
```
radu-mocanu: use empty/generic GUIDS in tests so we are not flagged by security reviews

PR #1739, opened 2026-06-22
feat(guardrails): send execution source and job key headers

valentinabojan:
Guardrail validation calls now send the execution **source** (`x-uipath-guardrails-source`) and **job key** (`x-uipath-jobkey`) headers for licensing/metering correlation.

The source is derived from the executing CLI command (`run`→`runtime`, `debug`/`dev`→`playground`, `eval`→`eval`) on `UiPathRuntimeContext.execution_source` (added in uipath-runtime 0.11.4), and flows through the execution-context boundary rather than an env var: the CLI calls `set_execution_source(ctx.execution_source)`, which sets a `ContextVar` that `UiPathExecutionContext.execution_source` reads, and `GuardrailsService` builds the header from `self._execution_context.execution_source`. This keeps the source correctly scoped for concurrent/eval runs and works for both coded and low-code agents, since both run through the `uipath` CLI. Job key continues to come from `UiPathConfig.job_key`; both headers are omitted when unset.

Requires `uipath-runtime>=0.11.4` (companion PR: UiPath/uipath-runtime-python#132).


ctiliescuuipath (approved):

apetraru-uipath (approved):

radu-mocanu (changes requested):
env vars are acceptable for process bootstrap/config coming from the outside world, but this PR uses an env var as an internal (same-process) message bus from the CLI layer to the platform client. this is not a clean design as it:
- makes the state global
- can t be scoped in async/concurrent eval runs
etc.

 we should make this execution metadata part of runtimecontext and pass it through the existing context boundary, as currently done for for command/job/trace-related execution state.

 A cleaner shape would be:
 ```

  - map `command -> execution_source` when building `UiPathRuntimeContext`
  - carry it as `runtime_context.execution_source`
  - propagate it into `UiPathExecutionContext` / the platform client context
  - have `GuardrailsService` read `self._execution_context.execution_source` when adding `x-uipath-guardrails-source`
```

File: packages/uipath/src/uipath/_cli/cli_debug.py [hunk 1]
```diff
@@ -122,14 +126,16 @@ def debug(
...
+                ctx = UiPathRuntimeContext.with_defaults(
                     input=input,
                     input_file=input_file,
                     output_file=output_file,
                     resume=resume,
                     trace_manager=trace_manager,
                     command="debug",
-                ) as ctx:
+                )
+                set_execution_source(ctx.execution_source)
```
radu-mocanu: this should be a context manager (like ctx or the resource overwrites). the context var token should be released when we exit the context

valentinabojan: Done

packages/uipath-platform/src/uipath/platform/common/_execution_context.py
```diff
@@ -1,7 +1,37 @@
...
+_execution_source: ContextVar[Optional[str]] = ContextVar(
+    "execution_source", default=None
+)
+
+
+class ExecutionSourceContext:
+    """Scope the execution source for the duration of a run.
+
+    Carries the source (e.g. ``runtime``/``playground``/``eval``) via a context
+    variable instead of a process-global env var, and releases it on exit so it
```
radu-mocanu: please remove the references to the old impl `instead of a process-global env var`

packages/uipath-platform/src/uipath/platform/common/_execution_context.py
```diff
@@ -1,7 +1,37 @@
...
+    stays correctly scoped in concurrent runs. The CLI enters this with
+    ``UiPathRuntimeContext.execution_source`` so platform clients can read it
+    via :attr:`UiPathExecutionContext.execution_source`.
+    """
+
+    def __init__(self, execution_source: Optional[str]) -> None:
+        self._execution_source = execution_source
+        self._token: Optional[Token[Optional[str]]] = None
```
radu-mocanu: nit: let s use the new typing hints please
```suggestion
        self._token: Token[str | None] | None = None
```

radu-mocanu (approved):

PR #1746, opened 2026-06-24
feat(platform): add guardrailName to guardrails validation payload

ctiliescuuipath:
## Summary
- Adds `guardrailName` field to the guardrails validation request payload so the server can identify which guardrail is being evaluated
- Bumps `uipath-platform` version to 0.1.74
- Updates tests to verify the new field is present and correct

## Test plan
- [x] Updated `test_guardrails_service.py` to assert `guardrailName` is included in the payload
- [ ] Run full test suite: `cd packages/uipath-platform && pytest`

### DIFF
--- a/packages/uipath/uv.lock
+++ b/packages/uipath/uv.lock
@@ -2691,7 +2691,7 @@ dev = [
 
 [[package]]
 name = "uipath-platform"
-version = "0.1.75"
+version = "0.1.76"
 source = { editable = "../uipath-platform" }
 dependencies = [
     { name = "httpx" },
--- a/packages/uipath/uv.lock
+++ b/packages/uipath/uv.lock
@@ -2691,7 +2691,7 @@ dev = [
 
 [[package]]
 name = "uipath-platform"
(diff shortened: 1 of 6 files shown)


packages/uipath/uv.lock
```diff
@@ -3,7 +3,7 @@ revision = 3
 requires-python = ">=3.11"
 
 [options]
-exclude-newer = "2026-06-21T14:10:18.2328398Z"
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
```
valentinabojan: Is this needed?

valentinabojan (approved):

packages/uipath/uv.lock
ctiliescuuipath: was produced on uv sync i see this line `exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.` in multiple py projects agents core

PR #1749, opened 2026-06-24
fix(remote-a2a): fall back to folder context when folder_path is missing

radugheo:
- `RemoteA2aService._retrieve_spec` now resolves the folder through a `_resolve_folder_key` helper that mirrors `McpService`: when no `folder_path` is provided it falls back to the `FolderContext` instead of raising `Cannot obtain folder_key without providing folder_path`.
- Enables A2A runs without the `UIPATH_FOLDER_PATH` env var (e.g. `uipath debug` with no `UIPATH_FOLDER_PATH` and no resource override)
- Adds A2A service tests for both the fallback and the explicit-folder paths.
- Bumps `uipath-platform` to 0.1.75 and `uipath` to 2.11.12.

### DIFF
--- a/packages/uipath-platform/pyproject.toml
+++ b/packages/uipath-platform/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-platform"
-version = "0.1.74"
+version = "0.1.75"
 description = "HTTP client library for programmatic access to UiPath Platform"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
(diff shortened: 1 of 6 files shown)


PopescuTudor (approved):

radu-mocanu (commented):
## Golden-rule review

Reviewed the remote-a2a folder fallback change against the golden rules. The version bumps, lock files, and tests are consistent with the rules. No golden-rule violations found.

No golden-rule violations found.

edis-uipath (approved):

PR #1753, opened 2026-06-24
fix(tracing): exclude a2a-sdk instrumentation scope from LLMOps export

PopescuTudor:
## Problem

The a2a-sdk auto-instruments its JSON-RPC transport — every method of `JsonRpcTransport` is wrapped under the OpenTelemetry instrumentation scope **`a2a-python-sdk`**, emitting spans like `a2a.client.transports.jsonrpc.JsonRpcTransport.send_message/_get_http_args/_send_request`. These render as **noisy, unparented internal-method nodes** in the execution trace and aren't meaningful execution spans.

Low-code agents already drop them, because the agent runtime exports through `FilteringSpanExporter(filter_fn=is_custom_instrumentation_span)` — and a2a-sdk spans aren't `uipath.custom_instrumentation`. But **coded agents run with no span filter** (`settings=None` → `span_filter=None`), so the a2a-sdk spans pass straight through to the trace.

## Change

Drop spans from excluded third-party instrumentation scopes **at the span-processor layer**, before and independent of the optional `UiPathTraceSettings.span_filter`, so the exclusion applies to every agent (coded *and* low-code), ordering-independent and with no global env mutation:

- **uipath-core** — add `EXCLUDED_INSTRUMENTATION_SCOPES = frozenset({"a2a-python-sdk"})` and `is_excluded_instrumentation_scope(span)` in `core/tracing/types.py`, and apply it in `UiPathExecutionTraceProcessorMixin.on_end` (batch/simple export path).
- **uipath** — apply the same check in `LiveTrackingSpanProcessor.on_start` / `on_end` (the live-upsert path that feeds the job Execution Trace).

The denylist is a single, extensible set, mirroring the existing `is_custom_instrumentation_span` filter convention.

## Tests

- `uipath-core`: `is_excluded_instrumentation_scope` matches a2a-sdk spans and ignores normal spans; a2a-sdk-scope spans are dropped even with `span_filter=None` (the coded path).
- `uipath`: `LiveTrackingSpanProcessor.on_start`/`on_end` skip excluded-scope spans.
- `ruff`, `ruff format`, `mypy`, and both packages' affected test modules pass locally (9/9 and 34/34).

## Versioning

Bumps `uipath-core` 0.5.22→0.5.23 and `uipath` 2.11.12→2.11.13, and raises `uipath`'s minimum on `uipath-core` to `>=0.5.23` (co-changed internal packages).

https://claude.ai/code/session_01ELDu3m5eaURJrMVarc8VfY

### DIFF
--- a/packages/uipath-core/src/uipath/core/tracing/processors.py
+++ b/packages/uipath-core/src/uipath/core/tracing/processors.py
@@ -11,7 +11,10 @@ from opentelemetry.sdk.trace.export import (
     SpanExporter,
 )
 
-from uipath.core.tracing.types import UiPathTraceSettings
+from uipath.core.tracing.types import (
+    UiPathTraceSettings,
+    is_excluded_instrumentation_scope,
+)
 
 
 class UiPathExecutionTraceProcessorMixin:
@@ -34,6 +37,10 @@ class UiPathExecutionTraceProcessorMixin:
 
     def on_end(self, span: ReadableSpan):
         """Called when a span ends. Filters before delegating to parent."""
+        # Always drop third-party instrumentation noise, then apply the
+        # optional span filter.
+        if is_excluded_instrumentation_scope(span):
+            return
         span_filter = self._settings.span_filter if self._settings else None
         if span_filter is None or span_filter(span):
             parent = cast(SpanProcessor, super())
--- a/packages/uipath-core/pyproject.toml
+++ b/packages/uipath-core/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-core"
-version = "0.5.22"
+version = "0.5.23"
 description = "UiPath Core abstractions"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/packages/uipath-core/src/uipath/core/tracing/__init__.py
+++ b/packages/uipath-core/src/uipath/core/tracing/__init__.py
@@ -7,11 +7,19 @@ with OpenTelemetry tracing, including custom processors for UiPath execution tra
 from uipath.core.tracing.decorators import traced
 from uipath.core.tracing.span_utils import UiPathSpanUtils
 from uipath.core.tracing.trace_manager import UiPathTraceManager
-from uipath.core.tracing.types import UiPathTraceSettings
+from uipath.core.tracing.types import (
+    EXCLUDE_SCOPES_FEATURE_FLAG,
+    EXCLUDED_INSTRUMENTATION_SCOPES,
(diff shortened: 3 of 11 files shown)


packages/uipath-core/src/uipath/core/tracing/processors.py
```diff
@@ -34,6 +37,10 @@ def on_start(self, span: Span, parent_context: context_api.Context | None = None
 
     def on_end(self, span: ReadableSpan):
         """Called when a span ends. Filters before delegating to parent."""
+        # Always drop third-party instrumentation noise, then apply the
+        # optional span filter.
+        if is_excluded_instrumentation_scope(span):
```
radu-mocanu, line 42: **[GR-4]** New span-exclusion behavior is enabled unconditionally

The new always-on drop of a2a-python-sdk spans runs for every agent (coded and low-code) with no kill switch. New behavior should default OFF behind an env-var feature flag (UIPATH_FEATURE_<NAME>) on a patch version, then flip the default on later.

**Fix:** Gate is_excluded_instrumentation_scope behind an env-var feature flag (default off), or document why this is a safe deterministic noise filter exempt from the flag policy.

radu-mocanu (changes requested):
## Golden-rule review

The change drops third-party a2a-sdk instrumentation noise at the span-processor layer in uipath-core and uipath. Versions are bumped in both packages and the uipath->uipath-core minimum is raised to 0.5.23, with uv.lock updated. The one golden-rule concern is that this new always-on exclusion behavior is enabled unconditionally rather than gated behind an opt-in feature flag.

- **[GR-4]** `packages/uipath-core/src/uipath/core/tracing/processors.py`:42: New span-exclusion behavior is enabled unconditionally
  - The new always-on drop of a2a-python-sdk spans runs for every agent (coded and low-code) with no kill switch. New behavior should default OFF behind an env-var feature flag (UIPATH_FEATURE_<NAME>) on a patch version, then flip the default on later.
  - Fix: Gate is_excluded_instrumentation_scope behind an env-var feature flag (default off), or document why this is a safe deterministic noise filter exempt from the flag policy.

_Automated review against the repo's golden rules. Reply if a rule is misapplied._

PopescuTudor:
Addressed **GR-4**: the exclusion is now gated behind the `UIPATH_FEATURE_ExcludeThirdPartyTraceScopes` feature flag (default **off**), via the existing `FeatureFlags` manager, so it ships dark and the default can be flipped on in a later release once validated. `is_excluded_instrumentation_scope` returns `False` unless the flag is enabled; added tests for both the flag-on (dropped) and flag-off (kept) paths.

PR #1754, opened 2026-06-24
fix(tracing): fall back to TRACE_REFERENCE_ID for local workspace spans

GabrielVasilescu04:
For local workspace runs, PROJECT_KEY is not set so agentId was never added to attributes_dict, causing reference_id to be None and bypassing the default_factory. Spans were stored with ReferenceId=NULL, breaking feedback lookup and trace association for design-time runs.

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/common/_span_utils.py
+++ b/packages/uipath-platform/src/uipath/platform/common/_span_utils.py
@@ -283,7 +283,7 @@ class _SpanUtils:
         # Top-level fields for internal tracing schema
         execution_type = attributes_dict.get("executionType")
         agent_version = attributes_dict.get("agentVersion")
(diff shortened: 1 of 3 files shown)


radu-mocanu (commented):
## Golden-rule review

The change adds a fallback to an environment variable for the span reference id and bumps the package version, with the lock file regenerated accordingly. The reference id and surrounding span fields already read this same environment variable through the dataclass default factory, so reading it here is consistent with existing behavior and not a new internal message bus. No golden rule violations found.

No golden-rule violations found.

---

> ### 🤖 UiPath Coded Agent review
> This pull request was reviewed automatically by a **UiPath Coded Agent**, built with the [UiPath Python SDK](https://github.com/UiPath/uipath-python) that lives in this very repository. It enforces only the repo's non-negotiable golden rules, nothing else.

PR #1758, opened 2026-06-24
rearrange spans for eval trace

PopescuTudor:
---

### DIFF
--- a/src/uipath/_cli/_evals/_runtime.py
+++ b/src/uipath/_cli/_evals/_runtime.py
@@ -18,6 +18,7 @@ from typing import (
 
 import coverage
 from opentelemetry import context as context_api
+from opentelemetry import trace
 from opentelemetry.sdk.trace import ReadableSpan, Span
 from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
 from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


src/uipath/_cli/_evals/_runtime.py
```diff
@@ -299,60 +300,72 @@ async def execute(self) -> UiPathRuntimeResult:
...
-                    evaluation_set_name=evaluation_set.name,
-                    evaluation_set_results=eval_run_result_list,
-                )
+                # Create the parent "Evaluation set run" span
+                tracer = trace.get_tracer(__name__)
+                span_attributes = {
+                    "execution.id": self.execution_id,
+                    "span_type": "eval_set_run",
+                }
+                if self.context.eval_set_run_id:
```
radu-mocanu, line 309: **No version bump visible for this logic change**

This change alters runtime behavior by adding and rearranging trace spans, which is a logic change that should come with a package version bump. The diff does not show a version bump. Please confirm a version bump exists, and if not, add one.

**Fix:** Bump the affected package version in the same change.

radu-mocanu (changes requested):
### Golden-rule review
*Produced by a UiPath Coded Agent, itself built with the [UiPath Python SDK](https://github.com/UiPath/uipath-python), the code that lives in this very repository.*

---

- `src/uipath/_cli/_evals/_runtime.py`:309: No version bump visible for this logic change
  - This change alters runtime behavior by adding and rearranging trace spans, which is a logic change that should come with a package version bump. The diff does not show a version bump. Please confirm a version bump exists, and if not, add one.
  - Fix: Bump the affected package version in the same change.

PR #1760, opened 2026-06-24
fix: conversational events - client side tool event responses should be queued

norman-le:
### Problem
When the LLM calls multiple tools in one turn, the bridge stored each `endToolCall`/`confirmToolCall` response in a single variable, overwriting previous values. If responses arrived before `wait_for_resume()` consumed them, earlier values were lost — causing the exchange to hang.

### Fix: `_bridge.py`

**Replaced single-slot storage with keyed dict matching using 3 data structures:**

1. **`_expected_tool_call_ids`** (deque) — ordered queue of tool_call_ids, populated by `emit_interrupt_event()`, consumed by `wait_for_resume()` to know which response to match
2. **`_tool_resume_results`** (dict) — responses that arrived before `wait_for_resume()` was called; checked first for immediate return
3. **`_tool_resume_pending`** (dict of Futures) — created by `wait_for_resume()` when response hasn't arrived yet; resolved by `_resolve_or_store_resume()` when it does
4. 
*Flow*: wait_for_resume pops the next expected ID from the deque → checks results (already arrived?) → if yes, return it → if no, create a pending Future → _resolve_or_store_resume resolves it when the response finally comes in.

**Key methods changed:**
- `emit_interrupt_event()` — now registers the trigger's `tool_call_id` in the deque (still no websocket emit, just using it here since it's naturally passed through with the known tool call ID)
- `wait_for_resume()` — pops expected ID from deque, checks results dict, falls back to Future
- `_resolve_or_store_resume()` — new helper that routes incoming responses to the correct Future or stores them
- `_handle_conversation_event()` — calls `_resolve_or_store_resume()` instead of setting a single variable

### Tests: `test_bridge.py`

- Fixed 3 existing tests to register expected tool_call_ids
- Added 5 concurrent scenario tests (all early, out of order, mixed, confirm+end)
- Added 8 edge case tests (empty deque, duplicates, malformed events, cleanup)
- 78 tests total, all passing

### Not changed
- `UiPathChatProtocol` — same interface
- `runtime.py` — untouched

Still works for client side tools, tool confirmation and server side tools:

https://github.com/user-attachments/assets/6d13c9a9-18ca-4871-b748-fdb06a8a1dd1

From our react-sdk, there's a natural gap, so to emulate what a client can do I buffered the client side tool responses to be received at the same time in the runtime/bridge before the resume function is called for the interrupts (lines 22-33 in my log file). Can see how the logs were used in the first commit of the PR: 53303a4f191919304167871e0d5421ad386dbcd7

https://github.com/user-attachments/assets/67833082-977e-46ba-86cc-bcf812fe26f2


packages/uipath/src/uipath/_cli/_chat/_bridge.py
```diff
@@ -410,21 +441,56 @@ async def emit_executing_tool_call_event(
...
+        if expected_id in self._tool_resume_results:
+            # Response arrived before we got here — return it immediately
+            item = self._tool_resume_results.pop(expected_id)
+        else:
+            # Response hasn't arrived yet — wait for it
+            future: asyncio.Future[ToolResumeItem] = (
+                asyncio.get_running_loop().create_future()
+            )
+            self._tool_resume_pending[expected_id] = future
+            item = await future
```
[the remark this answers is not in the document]
norman-le: Valid concern in theory, but doesn't happen in practice. The runtime always calls them in pairs:

```python
for trigger in api_triggers:
    await self.chat_bridge.emit_interrupt_event(trigger)  # register
    resume_data = await self.chat_bridge.wait_for_resume()  # consume
```

So the first `wait_for_resume()` always completes before the second `emit_interrupt_event()` registers the same ID again. The deque never has two of the same ID simultaneously.

For `requireConversationalConfirmation`, the confirm and end phases happen in **separate suspension rounds** — the delegate suspends once with confirm triggers, resumes, then suspends again with end triggers. Same sequential pairing.

Our test `test_confirm_then_end_same_tool_call` covers exactly this flow and passes because the confirm is consumed before the end is registered.

If someone ever changed the runtime to register all triggers up front before waiting, it would break. But that would be a runtime change, not a bridge issue. We could defensively use a list per key instead of a single entry, but it adds complexity for a scenario that can't occur with the current caller.

packages/uipath/src/uipath/_cli/_chat/_bridge.py
```diff
@@ -410,21 +441,77 @@ async def emit_executing_tool_call_event(
...
+
```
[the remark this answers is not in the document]
norman-le: When the websocket disconnects, the job either suspends or terminates on the CAS side. The bridge instance is discarded — all its state (deque, results, pending futures) gets garbage collected. No one will ever call `wait_for_resume()` on it again.

So a leaked Future or lost expected ID during shutdown has no impact — there's nothing left to consume them.

packages/uipath/src/uipath/_cli/_chat/_bridge.py
```diff
@@ -410,21 +441,77 @@ async def emit_executing_tool_call_event(
...
+        and returns the matching response.  Two cases:
+
+        1. Response already arrived (stored in _tool_resume_results) — return
+           immediately without blocking.
+        2. Response hasn't arrived yet — create a Future in _tool_resume_pending,
+           block until _handle_conversation_event resolves it.
+
+        Returns:
+            The resume data dict, including ``tool_call_id``.
+        """
```
maxduu, line 464: Nice! Thanks for the detailed flow + demoes.

Two questions:
- In the second demo, if you didn't submit the tool results in order (e.g. instead of 1,2,3,4,5 you did 3,5,1,4,2) would it work?
- More of a design question, let me know if I'm misunderstanding anything: Any reason we need to dequeue just the next expected tool_call_id? If we just had a data-structure list with all N tool-calls, could we simplify by whenever a tool-resume event occurs, checking all the tool-call-ids in that list to see matches?

maxduu (approved):
Overall I think I understand? But let's make sure to test all the edge-cases and left 2 questions [here](https://github.com/UiPath/uipath-python/pull/1760/changes#r3475165738)

Some other edge-cases I can think of:
- Parallel tool-calls with some client-side tools and others including/not-including tool-call confirmations (looks like you tested these already)
- Parallel tool-calls where some tools are Orchestrator process tools which do the suspend/resume - will your data-structures persist?

JoshParkSJ (approved):

packages/uipath/src/uipath/_cli/_chat/_bridge.py
norman-le: Good questions.

1. Yes, it works out of order. Responses get stored in `_tool_resume_results` by tool_call_id regardless of arrival order. When `wait_for_resume()` runs, it looks up by the expected ID, not by arrival order. The test `test_concurrent_tool_calls_out_of_order` tests this.


2. The deque exists because `wait_for_resume()` has no parameter — it doesn't know which tool_call_id to return. The runtime calls it N times in a sequential loop:
https://github.com/UiPath/uipath-runtime-python/blob/2b45546dbeb249541869a4fc227cac0839d51ca5/src/uipath/runtime/chat/runtime.py#L106

```python
for trigger in api_triggers:
    emit_interrupt_event(trigger)     # "I'm about to wait for toolcall-3"
    resume_data = wait_for_resume()   # "give me toolcall-3's result"
    resume_map[trigger.interrupt_id] = resume_data
```

Each call must return the result for *that specific trigger*. Without the deque, `wait_for_resume()` could return any available result and pair it with the wrong trigger (e.g. assign toolcall-1's result to toolcall-3, etc.)

One option was to pass `tool_call_id` as a parameter to `wait_for_resume(tool_call_id)` — then you'd only need the results/pending dicts, no deque. But that requires changing the `UiPathChatProtocol`, which I wanted to avoid. The deque bridges that gap without a protocol change.

maxduu: Makes sense, thanks!

PR #1772, opened 2026-06-30
refactor(eval)!: rename AgentExecution to WorkloadExecution

mjnovice:
## What changed?

Advances the unified-evals naming on the eval framework's central execution type, shipped as an **explicit breaking change** in a minor release.

1. **`AgentExecution` → `WorkloadExecution`**, and the `agent_execution` identifier (locals + the public `evaluate()` / `validate_and_evaluate_criteria()` parameter) → `workload_execution`.
2. **`WorkloadExecution.agent_trace` → `workload_trace`** and **`WorkloadExecution.agent_output` → `workload_output`** (the two model fields), with all constructions/accesses updated.
3. Comments, docstrings, help text and test prose across the eval framework moved from "agent execution" to "workload execution" terminology.

The value passed to every evaluator represents *a workload execution* (agent, process/orchestration, case management, …), not specifically an agent.

> Remaining `WorkloadExecution` fields `agent_input` / `expected_agent_behavior` are left as-is (out of scope for this pass).

## ⚠️ Breaking changes (ship as `2.12.0`)

Per [`release_policy.md`](packages/uipath/docs/release_policy.md), breaking changes to public interfaces land in a minor `X.Y.0`, preceded by an RC, with a migration guide. Versioned **`2.12.0`** (shipped directly as a stable minor, matching this repo's actual release practice — 615 published `uipath` versions, none of them RCs; past breaking minors `v2.9.0`/`v2.10.0` also shipped directly) and adds a [migration guide entry](packages/uipath/docs/core/release_notes.md).

- **`AgentExecution` → `WorkloadExecution`** — the class **name** is soft-deprecated via a module-level `__getattr__` shim (warns, resolves to `WorkloadExecution`); removed in **3.0**.
- **`WorkloadExecution.agent_trace`/`agent_output` → `workload_trace`/`workload_output`** — **hard break**. No field-level alias is provided (the type is in-process only); constructing/accessing the old field names raises `ValidationError` / `AttributeError`.
- **`evaluate()` / `validate_and_evaluate_criteria()` keyword `agent_execution` → `workload_execution`** — hard break for keyword callers. The internal runtime dispatch now calls **positionally**, so custom evaluators that still *declare* the old parameter name keep working; only callers passing `agent_execution=` by keyword are affected.

**No JSON/wire-contract change.** `WorkloadExecution` is in-process only. The identically-named `agent_output` / `agent_execution_time` fields on `EvalRunUpdatedEvent` and `StudioWebProgressItem` (a Studio Web wire contract serialized to camelCase) and the `AgentExecutionTimeMs` telemetry property are intentionally **left unchanged**.

## Review feedback addressed

- **[Chibionos] Release contract (patch → minor):** versioned as `2.12.0`; added release-notes row + migration guide.
- **[Chibionos] Custom-evaluator keyword compatibility:** internal dispatch now positional (`runtime.py`), so subclasses overriding with the old `agent_execution` parameter name still run. Covered by `test_old_param_name_still_dispatches_positionally`.
- **[Chibionos] Field compatibility:** per the maintainer decision to make this an explicit break, the renamed fields are **not** aliased (no `AliasChoices` / deprecated properties). Documented in the migration guide and locked by `test_old_field_names_are_a_hard_break`.
- **[Copilot] Duplicate `EvaluatorType` in `eval.models.__all__`:** removed.
- **[Copilot] Stale "agent execution" docstrings** (`output_evaluator.py`, `line_by_line_utils.py`): updated to "workload execution".

## Backward compatibility

`AgentExecution` (the name) keeps working with a `DeprecationWarning` via `__getattr__` in both `models/__init__.py` and `models/models.py`. It is intentionally **not** re-added to `__all__` — a soft-deprecated name should not be pulled in via `from ... import *`; explicit `from uipath.eval.models import AgentExecution` still works (with the warning).

## How has this been tested?

- `uv run pytest tests/evaluators tests/cli/eval` — **919 passing**
- `uv run mypy src/uipath/eval` — **Success: no issues found in 68 source files**
- `uv run ruff check . && uv run ruff format --check .` — **clean**
- New tests: class-name shim warns/resolves on both import paths, unknown-attr still raises, old field names hard-break, old parameter name dispatches positionally.


AAgnihotry (approved):

packages/uipath/src/uipath/eval/models/models.py
```diff
@@ -11,14 +12,14 @@
...
-    """Represents the execution data of an agent for evaluation purposes."""
+class WorkloadExecution(BaseModel):
+    """Represents the execution data of a workload for evaluation purposes."""
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     agent_input: dict[str, Any] | None
-    agent_output: dict[str, Any] | str
-    agent_trace: list[ReadableSpan]
+    workload_output: dict[str, Any] | str
```
Chibionos, line 21: **P1 - public API compatibility:** the `AgentExecution` shim preserves only the old class name, not the old model fields. I verified that `AgentExecution(agent_input={}, agent_output={"result": "ok"}, agent_trace=[])` now resolves to `WorkloadExecution` and raises a Pydantic `ValidationError` because `workload_output` and `workload_trace` are required. `from uipath.eval.models import *` also drops `AgentExecution`, and the shipped custom evaluator template still imports `AgentExecution` and reads `.agent_output` / `.agent_trace`.

If this is meant to be a deprecation-compatible patch, keep the old constructor keywords and attribute reads working until removal, for example with validation aliases and deprecated properties:

```python
from pydantic import AliasChoices, Field

class WorkloadExecution(BaseModel):
    model_config = ConfigDict(arbitrary_types_allowed=True, populate_by_name=True)

    agent_input: dict[str, Any] | None
    workload_output: dict[str, Any] | str = Field(
        validation_alias=AliasChoices("workload_output", "agent_output")
    )
    workload_trace: list[ReadableSpan] = Field(
        validation_alias=AliasChoices("workload_trace", "agent_trace")
    )

    @property
    def agent_output(self) -> dict[str, Any] | str:
        warnings.warn(
            "agent_output is deprecated and will be removed in uipath 3.0; use workload_output instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        return self.workload_output

    @property
    def agent_trace(self) -> list[ReadableSpan]:
        warnings.warn(
            "agent_trace is deprecated and will be removed in uipath 3.0; use workload_trace instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        return self.workload_trace
```

Also add `AgentExecution` back to `__all__` if wildcard imports remain part of the public compatibility surface.

packages/uipath/src/uipath/eval/runtime/runtime.py
```diff
@@ -1017,15 +1017,15 @@ async def run_evaluator(
...
-                agent_output=output_data,
-                agent_trace=execution_output.spans,
+                workload_output=output_data,
+                workload_trace=execution_output.spans,
                 expected_agent_behavior=eval_item.expected_agent_behavior,
             )
 
             result = await evaluator.validate_and_evaluate_criteria(
-                agent_execution=agent_execution,
+                workload_execution=workload_execution,
```
Chibionos: **P1 - custom evaluator compatibility:** this runtime call now passes `workload_execution=` as a keyword. Any custom evaluator that overrides `validate_and_evaluate_criteria(self, agent_execution, evaluation_criteria)` with the old public parameter name now fails with `TypeError: unexpected keyword argument 'workload_execution'` before it can run.

Because this is an internal dispatch site, use positional arguments here so both old and new parameter names continue to work:

```python
result = await evaluator.validate_and_evaluate_criteria(
    workload_execution,
    evaluation_criteria,
)
```

packages/uipath/src/uipath/eval/evaluators/base_evaluator.py
```diff
@@ -626,7 +626,7 @@ def model_post_init(self, __context: Any) -> None:
             self.description = self.evaluator_config.description
 
     async def validate_and_evaluate_criteria(
```
Chibionos, line 628: **P1 - public keyword compatibility:** the public evaluator API now declares `workload_execution`, so existing callers using `evaluator.validate_and_evaluate_criteria(agent_execution=...)` fail with `TypeError` before any deprecation path runs. The same issue applies to direct `evaluate(agent_execution=...)` calls on concrete evaluators.

If this is intended to be a soft deprecation, accept the old keyword as an alias for at least one minor cycle:

```python
def _resolve_execution_arg(
    workload_execution: WorkloadExecution | None,
    agent_execution: WorkloadExecution | None,
) -> WorkloadExecution:
    if workload_execution is None and agent_execution is not None:
        warnings.warn(
            "agent_execution is deprecated and will be removed in uipath 3.0; use workload_execution instead.",
            DeprecationWarning,
            stacklevel=3,
        )
        return agent_execution
    if workload_execution is None:
        raise TypeError("workload_execution is required")
    return workload_execution

async def validate_and_evaluate_criteria(
    self,
    workload_execution: WorkloadExecution | None = None,
    evaluation_criteria: Any = None,
    *,
    agent_execution: WorkloadExecution | None = None,
) -> EvaluationResult:
    workload_execution = _resolve_execution_arg(workload_execution, agent_execution)
    ...
```

For concrete `evaluate()` methods, either keep the old parameter name until the removal release or add the same aliasing pattern where keyword callers are supported.

packages/uipath/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.11.18"
+version = "2.11.19"
```
Chibionos: **P1 - release contract:** this PR still packages the rename as `2.11.19`, a patch release, while the PR description explicitly calls out hard breaks for the `evaluate()` keyword and `WorkloadExecution.agent_trace` / `agent_output` field access. The repo release policy says patch releases maintain backward compatibility; public breaking changes belong in an `X.Y.0` minor release with release notes and a migration guide.

Either add the compatibility shims so this is truly patch-safe, or ship it as the next minor/RC and document the migration:

```toml
[project]
name = "uipath"
version = "2.12.0rc1"
```

If the compatibility fixes land and the hard breaks go away, keeping `2.11.19` is fine, but the PR description should stop declaring breaking API changes.

Chibionos (changes requested):
Deep-review P1 findings only. These are public API compatibility and release-contract blockers; please fix before merge.

packages/uipath/src/uipath/eval/models/models.py
mjnovice: Addressed by treating this as an explicit breaking change (now shipped as `2.12.0`, a stable minor) rather than a patch. Per maintainer decision we are **not** adding field-level back-compat (no `AliasChoices`/deprecated properties): the renamed fields are an intentional hard break, documented in the migration guide and locked by `test_old_field_names_are_a_hard_break`. The class **name** stays soft-deprecated via the `__getattr__` shim. We intentionally keep `AgentExecution` out of `__all__` so it isn't pulled in by `import *` (explicit import still works, with the warning).

packages/uipath/src/uipath/eval/runtime/runtime.py
mjnovice: Done — the internal dispatch now passes positionally, so custom evaluators that still declare the old `agent_execution` parameter name keep working. Covered by `test_old_param_name_still_dispatches_positionally`.

packages/uipath/src/uipath/eval/evaluators/base_evaluator.py
mjnovice: Since this ships as an explicit breaking minor (`2.12.0`), we are not aliasing the old keyword — the `agent_execution` → `workload_execution` keyword rename is a documented break in the migration guide. We did make the internal dispatch positional (see runtime.py thread) so evaluator *subclasses* with the old parameter name still run; only callers passing `agent_execution=` by keyword are affected.

packages/uipath/pyproject.toml
mjnovice: Fixed — bumped to `2.12.0` (minor) and added a release-notes row plus a migration guide section documenting all three breaks. Shipping directly as a stable minor (not an RC) to match this repo's actual release practice — 615 published `uipath` versions, none of them RCs; past breaking minors v2.9.0/v2.10.0 also shipped directly.

Chibionos (approved):

PR #1776, opened 2026-06-30
feat(governance): non-blocking LiveTrackEventDispatcher for /runtime/log

viswa-uipath:
## Summary

The runtime's `TrackEventAuditSink` invokes its injected `track_event` callable synchronously on the agent's hook thread. Wiring `UiPathPlatformGovernanceProvider.track_event` directly would block agent execution on the `POST /runtime/log` HTTP round-trip.

`LiveTrackEventDispatcher` wraps `provider.track_event` in a bounded `ThreadPoolExecutor`. The host wires `dispatcher.dispatch` into `AuditManager(track_event=...)` instead — same kwargs, but the HTTP work runs on a background worker.

## What's in

- **`LiveTrackEventDispatcher`** (`packages/uipath-platform/src/uipath/platform/governance/_live_track_event_dispatcher.py`)
  - Bounded pool (default `max_workers=10`, mirrors `LiveTrackingSpanProcessor`)
  - `dispatch(*, event_name, data=None, operation_id=None)` — drop-in for the provider's `track_event` signature
  - Fire-and-forget contract: `dispatch` returns immediately; worker exceptions are caught and logged at debug
  - `shutdown(wait=True)` drains in-flight submissions; idempotent (safe under multiple atexit hooks)
- Exported on `uipath.platform.governance` alongside the existing `UiPathPlatformGovernanceProvider`
- **Tests** (7 cases): non-blocking dispatch proven via `threading.Event` sync, kwarg forwarding, default flow-through (`data=None`, `operation_id=None`), exception swallowing, concurrency under burst (20 submissions), `shutdown` drain semantics, `shutdown` idempotence
- **Bumps**: `uipath-platform 0.1.84 → 0.1.85` (new public class needs a fresh wheel — `0.1.84` is already on PyPI)

## Why a separate adapter (not built into the provider)

The provider's `track_event` is the **synchronous platform call**. Tests and other internal seams might legitimately want to block on it. Wrapping it in a thread pool by default would force every caller into a fire-and-forget shape and hide HTTP errors. Keeping the non-blocking concern as a separate class lets the host opt into it deliberately for the runtime-sink wiring path.

## Test plan

- [x] `uv run ruff check .` — clean
- [x] `uv run ruff format --check .` — clean
- [x] `uv run mypy src tests` — clean (205 source files)
- [x] `uv run pytest tests/services/test_live_track_event_dispatcher.py tests/services/test_governance_service.py tests/services/test_governance_provider.py` — 63 passed
- [x] `uv sync --locked` — no drift across all three workspace packages
- [x] PyPI version check: `0.1.85` not yet published; `0.1.84` is the current latest

### uipath-platform

```toml
[project]
dependencies = [
  "uipath-platform==0.1.88.dev1017767069",

  # "uipath-platform>=0.1.88.dev1017760000,<0.1.88.dev1017770000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```

### uipath

```toml
[project]
dependencies = [
  "uipath==2.12.2.dev1017767069",

  # "uipath>=2.12.2.dev1017760000,<2.12.2.dev1017770000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]

[tool.uv]
override-dependencies = ["uipath-platform==0.1.88.dev1017767069"]
```

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/governance/_live_track_event_dispatcher.py
+++ b/packages/uipath-platform/src/uipath/platform/governance/_live_track_event_dispatcher.py
@@ -1,0 +1,287 @@
+"""Non-blocking dispatcher for governance track-event telemetry.
+
+Wraps :meth:`UiPathPlatformGovernanceProvider.track_event_async` on a
+private background ``asyncio`` event loop so sync callers can fire
+telemetry events without blocking on the underlying ``POST /runtime/log``
+HTTP round-trip.
+
+:meth:`LiveTrackEventDispatcher.dispatch` is a sync fire-and-forget
+method that mirrors the kwargs of ``track_event_async``. Internally it
+schedules the async HTTP call onto a dedicated background loop, so the
+calling thread never blocks on network I/O and the underlying HTTP call
+remains async end-to-end.
+
+Design notes:
+
+- **Async HTTP inside, sync interface outside.** ``dispatch`` is a
+  sync function. Internally it enqueues a coroutine that awaits
+  ``provider.track_event_async``.
+
+- **Loop affinity.** ``httpx.AsyncClient`` lazy-binds its connection
+  pool to the first event loop that awaits on it. This dispatcher
+  assumes it owns the provider's async HTTP path — nothing else in
+  the process should await ``track_event_async`` (or any other
+  ``*_async`` method on the same underlying service) on a *different*
+  loop. See "one dispatcher per provider" below.
+
+- **Backpressure.** A ``BoundedSemaphore`` caps in-flight coroutines;
+  submissions that exceed the cap are dropped with a warning so
+  memory stays bounded when the backend is slow.
+
+- **Fire-and-forget contract.** Coroutine exceptions are observed on
+  the returned ``concurrent.futures.Future`` (to suppress asyncio's
+  "exception was never retrieved" warning) and logged at debug — they
+  cannot reach the caller because ``dispatch`` returns before the
+  coroutine runs.
+
+One dispatcher per provider. The dispatcher's background loop must be
+the only loop that awaits the provider's async methods.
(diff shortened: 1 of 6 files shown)


radu-mocanu (approved):
changes look good but please follow-up with the async variant as well. we should not rely on synchronous API calls

packages/uipath-platform/src/uipath/platform/governance/_live_track_event_dispatcher.py
```diff
@@ -0,0 +1,293 @@
...
+        Python 3.8+), which is *not* caught by
+        ``Future._invoke_callbacks``'s ``except Exception`` and would
+        propagate through the loop-thread's callback machinery during
+        shutdown. ``exception()`` simply returns the exception object
+        (or ``None``), which is enough to suppress asyncio's
+        "exception was never retrieved" warning at GC time. ``_run``
+        already logs the underlying failure; this callback exists
+        purely for accounting.
+        """
+        future.exception()  # observe to suppress GC-time warnings
```
radu-mocanu: This callback still fails on the cancellation path. `run_coroutine_threadsafe(...)` returns a `concurrent.futures.Future`, and for that type `future.exception()` raises `concurrent.futures.CancelledError` when the future was cancelled. So during `shutdown(wait=False)`, this line can raise before the future is removed from `_futures` and before the in-flight semaphore is released.

Please make the accounting run in a `finally`, e.g.:

```python
try:
    future.exception()
except concurrent.futures.CancelledError:
    pass
finally:
    with self._futures_lock:
        self._futures.discard(future)
    self._inflight.release()
```

This keeps the callback safe for success, failure, and cancellation.

PR #1800, opened 2026-07-07
fix: shutdown trace manager, not the executor, between jobs

andreiancuta-uipath:
Regression introduced [here](https://github.com/UiPath/uipath-python/blame/53c894beab91afebdd3ee11d2ee453c9f774fb04/packages/uipath/src/uipath/_cli/cli_run.py#L328).

`LiveTrackingSpanProcessor.shutdown()` only shuts down its internal ThreadPoolExecutor. It does not remove the processor from OpenTelemetry’s process-wide processor list. After that, the processor is dead but still subscribed to span start/end events, so the next job’s span is still delivered to job 1’s processor. Since that is shutdown, we get an error:
  `RuntimeError: cannot schedule new futures after shutdown`

```mermaid
sequenceDiagram
    participant Server as cli_server.handle_start
    participant Run as cli_run.run
    participant TM as UiPathTraceManager
    participant DSP as _DelegatingSpanProcessor
    participant Live1 as job 1 LiveTrackingSpanProcessor
    participant Exec1 as job 1 ThreadPoolExecutor
    participant Live2 as job 2 LiveTrackingSpanProcessor
    participant Exec2 as job 2 ThreadPoolExecutor

    Server->>Run: job 1 via asyncio.to_thread(cmd.main, ...)
    Run->>TM: create trace manager
    Run->>Live1: create live tracking processor
    Run->>TM: add_span_processor(live1)
    TM->>DSP: add(live1)
    Run->>DSP: traced job 1 span starts
    DSP->>Live1: on_start(span)
    Live1->>Exec1: submit(upsert_span)
    Run->>Live1: live1.shutdown()
    Live1->>Exec1: shutdown(wait=True)
    Note over DSP,Live1: live1 is shut down but still registered

    Server->>Run: job 2 in the same process
    Run->>TM: create new trace manager
    Run->>Live2: create live tracking processor
    Run->>TM: add_span_processor(live2)
    TM->>DSP: add(live2)
    Run->>DSP: traced job 2 first span starts
    DSP->>Live1: on_start(span)
    Live1->>Exec1: submit(upsert_span)
    Exec1-->>Live1: RuntimeError: cannot schedule new futures after shutdown
    Live1-->>DSP: exception propagates
    DSP-->>Run: span start fails before job 2 code runs
    Run-->>Server: result file is written as faulted
    Note over Live2,Exec2: live2 would be able to send job 2 spans, but live1 crashes first
```

Fix:
- Job-local span processors no longer leak into later jobs in the same uipath server process.
- Other trace processors registered by that job are also cleaned up, not just live tracking.
- Spans emitted during runtime/factory disposal still have a live processor available.
- New: also shutdown the LlmHttpOpsExporter from the LiveTrackingProcessor
- Reproduced issue locally, with `uipath server` and 2 agent jobs, and validated the fix the same way.
- Added an integration test; verified it catches the original regression.

### DIFF
--- a/packages/uipath/pyproject.toml
+++ b/packages/uipath/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.13.6"
+version = "2.13.7"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/packages/uipath/src/uipath/_cli/cli_debug.py
+++ b/packages/uipath/src/uipath/_cli/cli_debug.py
@@ -138,7 +138,6 @@ def debug(
                 with ExecutionSourceContext(ctx.execution_source), ctx:
                     factory: UiPathRuntimeFactoryProtocol | None = None
                     governance_bootstrap: GovernanceBootstrap | None = None
-                    live_tracking_processor: LiveTrackingSpanProcessor | None = None
 
                     try:
                         trigger_poll_interval: float = 5.0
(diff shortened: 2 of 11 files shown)


andreiancuta-uipath:
Sonar coverage complains about an exception branch that logs. ~~Not an essential path~~
Added a test.

radu-mocanu (approved):

PR #1820, opened 2026-07-17
fix: allow first push to a project with uninitialized file system

cristipufu:
## Problem

`uipath push` fails on a freshly created **Function** project with:

> :x: The targeted Studio Web project is not of type coded agent. Please check the UIPATH_PROJECT_ID environment variable.

or, depending on the backend response, an unhandled `EnrichedException` traceback (404).

The Studio Web backend returns **404** from `FileOperations/Structure` for projects whose file system was never initialized — even though the project exists (`GET /Project/{id}` returns 200 with `projectType: "Function"` and an empty `rootFolder`). Since `push` is exactly what would upload the first files, this is a bootstrap deadlock: the coded-agent validation (added in #937) blocks the very push that would initialize the project.

Verified against a live tenant: the FileOperations **write** path works fine on such a project; only `Structure` 404s until the first file exists. After the first push, `Structure` returns 200 with the full tree.

## Fix

- `StudioClient.get_project_structure_async`: treat a 404 from `FileOperations/Structure` as an **empty project structure** (first push bootstraps the files). Other HTTP errors propagate unchanged.
- `ensure_coded_agent_project_async`: an empty remote structure (no files, no folders) is a never-pushed project — allow it instead of raising `NonCodedAgentProjectException`.

## Testing

- New test `test_first_push_to_uninitialized_project`: Structure returns 404 → push succeeds and uploads all files.
- Existing tests unchanged: `test_push_non_coded_agent_project` (remote with files but no `pyproject.toml` still errors) and `test_push_with_api_error` (401 still propagates) keep passing.
- `tests/cli/test_push.py`: 33 passed; `tests/cli/test_pull.py`: 12 passed; ruff + mypy clean.

### DIFF
--- a/packages/uipath/pyproject.toml
+++ b/packages/uipath/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.13.11"
+version = "2.13.12"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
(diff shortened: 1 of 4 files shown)


andreitava-uip (approved):
should be fine. I don't think we have any callers that explicitly handle 404s

PR #1839, opened 2026-08-03
fix: categorize missing HITL index as deployment

radu-mocanu:
## Summary

- categorize missing Context Grounding indexes during DeepRAG and batch RAG trigger creation as deployment errors
- preserve already-classified trigger errors and the original missing-index detail
- guide users to check index deployment and folder configuration

## Why

Missing indexes were being caught by the generic HITL wrapper and reported as system errors. The Context Grounding handlers now own this classification.

### DIFF
--- a/packages/uipath-platform/src/uipath/platform/resume_triggers/_protocol.py
+++ b/packages/uipath-platform/src/uipath/platform/resume_triggers/_protocol.py
@@ -60,6 +60,7 @@ from uipath.platform.context_grounding.context_grounding_index import (
 from uipath.platform.errors import (
     BatchTransformFailedException,
     BatchTransformNotCompleteException,
+    ContextGroundingIndexNotFoundError,
     OperationNotCompleteException,
 )
 from uipath.platform.orchestrator.job import JobState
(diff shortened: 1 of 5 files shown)


packages/uipath-platform/src/uipath/platform/resume_triggers/_protocol.py
```diff
@@ -778,27 +781,34 @@ async def _handle_batch_rag_job_trigger(
...
+                        enable_web_search_grounding=value.enable_web_search_grounding,
+                        folder_path=value.index_folder_path,
+                        folder_key=value.index_folder_key,
+                    )
+            except ContextGroundingIndexNotFoundError as e:
+                raise UiPathFaultedTriggerError(
+                    ErrorCategory.DEPLOYMENT,
+                    "Context grounding index not found",
+                    str(e),
+                ) from e
```
andreitava-uip, line 819: nit: a deployment error should ideally be actionable by the user/org admin. We should add something in the message to point them in the right direction, like "Check if the index is deployed"

andreitava-uip (approved):

PR #1864, opened 2026-08-18
fix(cli): honor packOptions.filesIncluded for dot-prefixed paths

### LINKED ISSUE #1855: packOptions.filesIncluded is silently ignored for any path under a dot-prefixed directory
## What happens

`packOptions.filesIncluded` accepts a relative path so a file in a subdirectory can be added to the package. Any path under a dot-prefixed directory is dropped instead, and nothing reports it. The same applies to a dotfile named explicitly.

The spec describes `filesIncluded` as "Specific files to **always** include" ([packOptions](https://github.com/UiPath/uipath-python/blob/main/packages/uipath/specs/uipath.spec.md#4-packoptions)), and `uipath pack` exits successfully, so there is no signal that part of the configuration was ignored.

## Reproduction

```python
import os, tempfile
from uipath._cli._utils._project_files import files_to_include
from uipath._cli.models.uipath_json_schema import PackOptions

root = tempfile.mkdtemp()
os.makedirs(f"{root}/.config/assets/demo")
os.makedirs(f"{root}/plain")
open(f"{root}/.config/assets/demo/ASSET.md", "w").write("x")
open(f"{root}/plain/ASSET.md", "w").write("x")          # control, same shape without the dot
open(f"{root}/.python-version", "w").write("3.11")
open(f"{root}/pyproject.toml", "w").write("[project]")

opts = PackOptions(**{"filesIncluded": [
    ".config/assets/demo/ASSET.md",
    ".python-version",
    "plain/ASSET.md",
]})

included, skipped = files_to_include(opts, root, include_uv_lock=False)
print("included:", sorted(f.relative_path for f in included))
print("skipped :", skipped)
```

Output:

```
included: ['plain/ASSET.md', 'pyproject.toml']
skipped : []
```

`plain/ASSET.md` is honoured. `.config/assets/demo/ASSET.md` and `.python-version` are not, and `skipped` is empty.

## Cause

Two unconditional filters run before the inclusion rules are evaluated.

[`_project_files.py:471`](https://github.com/UiPath/uipath-python/blob/main/packages/uipath/src/uipath/_cli/_utils/_project_files.py#L471) drops dot-prefixed directories from `dirs[:]`, so `os.walk` never descends into them. [Line 497](https://github.com/UiPath/uipath-python/blob/main/packages/uipath/src/uipath/_cli/_utils/_project_files.py#L497) skips dot-prefixed files with a bare `continue`.

The per-file inclusion check that would honour the configuration is [lines 514 to 520](https://github.com/UiPath/uipath-python/blob/main/packages/uipath/src/uipath/_cli/_utils/_project_files.py#L514-L520), and it already supports a path match for subdirectories. It is simply unreachable for these paths.

Because both filters `continue` rather than appending to `skipped_files`, the omission does not surface anywhere, including in the caller that reports skipped files.

## Why it matters

Tooling configuration and assets conventionally live in a dot-prefixed directory. A project that needs such a directory inside its package has no way to express that: `directoriesExcluded` has no `directoriesIncluded` counterpart, and the documented per-file escape hatch does not work there.

The available workaround is to keep the content under a non-dot directory and copy it into place at runtime, which means the package layout no longer matches the source layout and the copy step has to be reimplemented per project.

## Suggested fix

Any of these would resolve it, in rough order of smallest change:

1. Honour an explicit `filesIncluded` entry regardless of a dot prefix. Descend into a dot-prefixed directory only when some `filesIncluded` entry has it as a path prefix, which keeps the current pruning fast for `.git`, `.venv` and similar. Apply the same to the dotfile filter at line 497. This matches the word "always" already in the spec.
2. Add `directoriesIncluded` to `packOptions` as the counterpart to `directoriesExcluded`.

Independent of which is chosen, appending to `skipped_files` at both filters would turn a silent omission into something a caller can report.

## Environment

- `uipath` 2.14.2, resolved from PyPI
- Python 3.14, macOS
- Line numbers above are against `main`

jaideeppyne:
## Description

Fixes #1855.

`files_to_include` (`_cli/_utils/_project_files.py`) is meant to honor `packOptions.filesIncluded` — documented as *"Specific files to always include."* But it prunes and skips dot-prefixed paths **before** the inclusion logic runs:

- the directory loop does `if d.startswith(".") or is_venv_dir(...): continue`, so `os.walk` never descends into e.g. `.config/`;
- the file loop does `if file.startswith("."): continue`, so dotfiles never reach the inclusion check.

As a result, an explicit include living under a dot-directory (e.g. `.config/assets/x.md`) or a named dotfile (e.g. `.python-version`) is **silently dropped** — `uipath pack` exits `0` with those always-include files missing and an empty `skipped` list (unreportable data loss in the package).

Repro on `main` (the issue's script): `included: ['plain/ASSET.md', 'pyproject.toml']`, `skipped: []` — `.config/assets/demo/ASSET.md` and `.python-version` are gone.

## Fix

- The walk now descends into a dot/venv directory **only when** a `files_included` entry lives under it (so `.git`/`.venv` stay pruned, but an explicitly-included subtree is reachable).
- A hidden file is packed **only when** it's explicitly listed by name or relative path — never merely because its extension matches `fileExtensionsIncluded`.

Default behavior is unchanged: dot/venv directories are still pruned and hidden files are still excluded unless explicitly requested. I deliberately kept the diff focused — honoring explicit includes closes the silent-data-loss path without flooding `skipped` with every `.git` entry (an independent idea floated in the issue).

## Tests

Added `TestFilesToIncludeExplicitHiddenPaths`: packs an explicit include inside `.config/assets/` and a `.python-version` dotfile, and asserts non-listed hidden paths (`.secret.md`, `.git/config.md`) stay excluded. It **fails on current `main`** (`assert '.config/assets/asset.md' in set()`) and passes with this change. The two existing hidden-file-exclusion tests still pass — no regression to the default.

I used a coding agent to help write and verify this change.

### DIFF
--- a/packages/uipath/src/uipath/_cli/_utils/_project_files.py
+++ b/packages/uipath/src/uipath/_cli/_utils/_project_files.py
@@ -465,16 +465,25 @@ def files_to_include(
         normalized_root_rel_path = root_rel_path.replace(os.sep, "/")
         is_root_evals_folder = normalized_root_rel_path == LEGACY_EVAL_FOLDER
 
-        # Skip all directories that start with . or are a venv or are excluded
+        # Skip directories that start with . or are a venv or are excluded,
+        # unless an explicit `files_included` entry lives under them.
         included_dirs = []
         for d in dirs:
-            if d.startswith(".") or is_venv_dir(os.path.join(root, d)):
-                continue
-
-            # Check if directory should be excluded
             dir_path = os.path.join(root, d)
             dir_rel_path = os.path.relpath(dir_path, directory)
             normalized_dir_rel_path = dir_rel_path.replace(os.sep, "/")
+
+            if d.startswith(".") or is_venv_dir(dir_path):
+                # Dot/venv directories are pruned by default, but still descend
+                # when an always-include path lives under them so that
+                # `packOptions.filesIncluded` entries are honored.
+                has_included_descendant = any(
+                    included == normalized_dir_rel_path
+                    or included.startswith(normalized_dir_rel_path + "/")
+                    for included in files_included
+                )
+                if not has_included_descendant:
+                    continue
 
             # Check exclusion: by dirname (for root level) or by relative path
             should_exclude_dir = (
@@ -494,9 +503,6 @@ def files_to_include(
 
         dirs[:] = included_dirs
         for file in files:
-            if file.startswith("."):
-                continue
-
             file_extension = os.path.splitext(file)[1].lower()
             file_path = os.path.join(root, file)
             file_name = os.path.basename(file_path)
@@ -505,20 +511,27 @@ def files_to_include(
             # Normalize the path
             normalized_rel_path = rel_path.replace(os.sep, "/")
 
+            # A file is explicitly requested when it matches a `files_included`
+            # entry by filename (base directory only) or by relative path.
+            explicitly_included = (
+                file in files_included and normalized_rel_path == file
+            ) or normalized_rel_path in files_included
+
+            # Hidden files are excluded by default, but an explicit
+            # `packOptions.filesIncluded` entry must still be honored.
+            if file.startswith(".") and not explicitly_included:
+                continue
+
             # Skip files in the root evals folder (but allow eval-set and evaluators subdirectories)
             if is_root_evals_folder:
                 skipped_files.append(normalized_rel_path)
                 continue
 
-            # Check inclusion: by extension, by filename (for base directory), or by relative path
+            # Check inclusion: by extension (visible files only), or when the
+            # file is explicitly listed by filename / relative path.
             should_include = (
-                file_extension in file_extensions_included
-                or (
-                    file in files_included and normalized_rel_path == file
-                )  # filename match for base directory only
-                or normalized_rel_path
-                in files_included  # path match for subdirectories
-            )
+                not file.startswith(".") and file_extension in file_extensions_included
+            ) or explicitly_included
 
             # Check exclusion: by filename (for base directory only) or by relative path
             should_exclude = (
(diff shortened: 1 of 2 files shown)


jaideeppyne:
Good catch, thanks — fixed in f3dd86f. Descending into a dot/venv directory for an explicit include no longer lets *other* visible files there be picked up by extension: I track whether the current directory is under a pruned (dot/venv) ancestor and, in that case, include a file only when it's explicitly listed in `filesIncluded` (never by extension). Extended the regression test with a visible, non-listed `.config/assets/other.md` and asserted it stays excluded.

jaideeppyne:
Thanks for the careful review — all three points are addressed (see 4cc5e0a on top of f3dd86f):

**1. [P1 / codex + Copilot] Extension-inclusion leaking non-listed files under a descended-into dot/venv dir.** Traversal still descends into a hidden/venv directory only when a `filesIncluded` entry lives beneath it, but inclusion there is now restricted to the explicitly-listed files. Each walk root carries a `root_under_pruned_dir` flag (set when any ancestor between it and the project root is dot-prefixed or a venv), and extension-based inclusion is gated on `not file.startswith(".") and not root_under_pruned_dir`. So an unlisted `.config/assets/other.md`, `.config/secrets.json`, or `.venv/.../tool.py` is no longer packed by extension — only the requested files (and the ancestor dirs needed to reach them) are admitted.

**2. [Copilot] Redundant `explicitly_included` predicate.** Collapsed to a single `normalized_rel_path in files_included`. Root-level entries are bare filenames where `normalized_rel_path == file`, so the old left operand was fully subsumed by the right; behavior is identical for both root-level and nested explicit includes.

**3. [Copilot] Test didn't assert non-listed siblings stay excluded.** The regression test now creates a *visible* `.config/assets/other.md` that is not in `filesIncluded` and asserts `".config/assets/other.md" not in rel_paths`. Verified fail-without/pass-with: reverting the `root_under_pruned_dir` guard makes exactly that assertion fail (`other.md` leaks in), and it passes with the fix.

`ruff check`, `ruff format --check`, `mypy`, and the full `tests/cli/` suite (1381 passed) are green.

PR #1866, opened 2026-08-19
feat: forward EntryPointPath when starting a process job

UiPathPetruPopa:
## What changed?

Two packages, one capability: a process tool can name **which entry point** of the referenced process it invokes.

- **`packages/uipath`** — `AgentProcessToolProperties` gains `entry_point_path` (alias `entryPointPath`).
- **`packages/uipath-platform`** — `ProcessesService` forwards it to Orchestrator as `StartInfo.EntryPointPath`.

**Inert until something sets it.** With the field absent the payload is byte-identical to today's, and Orchestrator resolves the release's configured entry point exactly as it does now. Nothing writes the field yet — the picker that does is gated behind its own flag in the Agents repo and lands after this.

### Two things that would have failed silently

**`**kwargs` would not have worked.** `invoke`/`invoke_async` accept `**kwargs`, but `_invoke_spec` is called with an explicit argument list and pulls exactly one name out of it (`parent_span_id`). An `entry_point_path` kwarg would have been accepted at the call site, dropped before the payload, and reported as success. It is threaded explicitly through all three.

**The key must be the literal `"EntryPointPath"`.** This payload is a plain dict of hardcoded PascalCase keys — no alias generator, no conversion layer — so a snake_case name would ship as-is. The model needs the same care for the mirror reason: `BaseCfg` sets `validate_by_name` and `validate_by_alias` but registers no alias generator, so every camelCase alias in that file is hand-written.

**Empty is omitted, not sent.** `""` would make Orchestrator resolve an entry point named `""`, which fails the release lookup; an absent key selects the configured default.

## How has this been tested?

- `tests/services/test_processes_service.py` — 11 tests pass. Two added: the path appears in `startInfo` when given, and an empty string is omitted.
- **The omission test was mutation-checked**: weakening the guard from `if entry_point_path:` to `if entry_point_path is not None:` makes it fail. It guards the behaviour rather than merely covering the line.
- The existing `test_invoke` asserts the exact payload and still passes unchanged — direct evidence the default path is untouched.
- `packages/uipath`: full model/agent suite green; `ruff check` and `ruff format --check` clean on both packages.

## Are there any breaking changes?
- [ ] Under Feature Flag
- [x] None
- [ ] DB migrations
- [ ] API removals
- [ ] Authentication/Authorization
- [ ] Deployment pipeline
- [ ] Infrastructure

Purely additive: a new optional model field and a new optional keyword argument.

## Order

Both packages must publish before `uipath-langchain` can raise its lower bounds and forward the stored value, and all of that precedes the picker UI — which is what makes the field user-reachable.


packages/uipath/src/uipath/agent/models/agent.py
```diff
@@ -948,6 +948,9 @@ class AgentProcessToolProperties(BaseResourceProperties):
 
     folder_path: Optional[str] = Field(None, alias="folderPath")
     process_name: Optional[str] = Field(None, alias="processName")
+    # Absent is a real mode: Orchestrator resolves the process's configured entry point when
+    # StartJobs omits EntryPointPath.
```
andreiancuta-uipath: nit: drop this comment; the one in _processes_service is sufficient

andreiancuta-uipath (approved):

packages/uipath/src/uipath/agent/models/agent.py
UiPathPetruPopa: Dropped in `e26a33bf`.

packages/uipath-platform/src/uipath/platform/orchestrator/_processes_service.py
```diff
@@ -52,6 +52,7 @@ def invoke(
         attachments: Optional[list[Attachment]] = None,
         parent_operation_id: Optional[str] = None,
         run_as_me: Optional[bool] = None,
+        entry_point_path: Optional[str] = None,
```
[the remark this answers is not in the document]
UiPathPetruPopa: Documented on both overloads in 8fd6a7e9, including what omitting it does — the key is left out of the request and Orchestrator resolves the release's configured entry point.

PR #1869, opened 2026-08-25
fix(cli): derive uipath pin in new scaffold from installed version

andreibalas-uipath:
`uipath new` scaffolded a `pyproject.toml` with a hard-coded `uipath>=2.10.0, <2.11.0` dependency that hadn't been bumped since 2.10, so the very next documented step (`uv sync`) downgraded the project's venv to 2.10.x on machines running 2.14.x (UiPath Jira [UV-16117](https://uipath.atlassian.net/browse/UV-16117), Issue 1).

**The fix:** `generate_pyproject()` now derives the pin from the installed distribution (via the existing `_get_safe_version()`), keeping the same `>=X.Y.0, <X.(Y+1).0` shape — pre-release/local suffixes are stripped — so uv resolution behavior is unchanged and only the numbers move. For every normal install (pip/uv/`uv tool install`) this is the path that runs.

**The fallback:** only when package metadata is unavailable (`_get_safe_version()` returns `"unknown"`, e.g. a source checkout run without installing) does it fall back to a literal range generated from a single constant, `FALLBACK_UIPATH_MINOR = "2.14"`, and print a warning. `test_fallback_pin_admits_installed_uipath` runs against the real installed package in CI and asserts the fallback admits the current release, so a future release PR that forgets to bump the constant fails CI instead of silently rotting. A second regression guard asserts the pin actually written by `uipath new` contains the installed `uipath` version.

**Also in the scaffold path:** the `Initialize project` hint's stray leading space is gone (it stays `uipath init` — with the pin derived from the running install, no sync step is needed before `init`). `uipath.json` now includes a GUID `id` (using the same `resolve_existing_project_id(...) or uuid4()` expression `init` uses for backfill), so `new` no longer emits `'id' field not present in uipath.json` right after creating the file. `packaging` is declared as a dev dependency (the tests import it; previously only transitive via pytest), and `docs/core/agents.md` no longer shows `uipath-langchain>=2.0` — that package is 0.x, so the spec resolved to nothing.

Package version bumped 2.14.8 → 2.14.9 per repo convention (every PR touching `packages/uipath` bumps the patch version; 2.14.8 was released by #1870 while this PR was open).

Follow-up (separate repos): the plugin CLIs (`uipath-langchain`, `uipath-llamaindex`, `uipath-openai-agents`) have the same stale-pin pattern; each should derive from its own installed version with its own minor as fallback and the same guard test.

[UV-16117]: https://uipath.atlassian.net/browse/UV-16117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/uipath/docs/core/agents.md
+++ b/packages/uipath/docs/core/agents.md
@@ -119,7 +119,7 @@ version = "0.1.0"
 description = "..."
 authors = [{ name = "Your Name", email = "you@example.com" }]
 requires-python = ">=3.11"
-dependencies = ["uipath>=2.0", "uipath-langchain>=2.0"]
+dependencies = ["uipath>=2.0", "uipath-langchain>=0.16"]
 ```
 
 Standard metadata plus the framework dependency (`uipath-langchain` here). The framework graph file and this dependency identify the project as a coded agent — `pyproject.toml` needs no UiPath-specific entries, and `uipath.json` carries no agent entry.
--- a/packages/uipath/src/uipath/_cli/cli_new.py
+++ b/packages/uipath/src/uipath/_cli/cli_new.py
@@ -1,6 +1,8 @@
 import json
 import os
+import re
 import shutil
+import uuid
 
 import click
 
@@ -8,9 +10,38 @@ from uipath.platform.constants import PYTHON_CONFIGURATION_FILE, UIPATH_CONFIG_F
 
 from ._telemetry import track_command
 from ._utils._console import ConsoleLogger
+from ._utils._project_files import resolve_existing_project_id
 from .middlewares import Middlewares
 
 console = ConsoleLogger()
+
+FALLBACK_UIPATH_MINOR = "2.14"
+
+
+def _minor_range_spec(major: int, minor: int) -> str:
+    return f"uipath>={major}.{minor}.0, <{major}.{minor + 1}.0"
+
+
+def _fallback_uipath_dependency_spec() -> str:
+    major, minor = (int(part) for part in FALLBACK_UIPATH_MINOR.split("."))
+    return _minor_range_spec(major, minor)
+
+
+def _uipath_dependency_spec() -> str:
+    from . import _get_safe_version
+
+    installed = _get_safe_version()
+    # Strip pre-release/dev/local suffixes: only the leading "major.minor" matters.
+    match = re.match(r"^(\d+)\.(\d+)", installed)
+    if match is None:
+        fallback = _fallback_uipath_dependency_spec()
+        console.warning(
+            f"Could not determine the installed 'uipath' version ('{installed}'); "
+            f"falling back to '{fallback}'. Pin it manually if needed."
+        )
+        return fallback
+
+    return _minor_range_spec(int(match.group(1)), int(match.group(2)))
 
 
 def generate_script(target_directory):
@@ -30,7 +61,7 @@ version = "0.0.1"
 description = "{project_name}"
 authors = [{{ name = "John Doe", email = "john.doe@myemail.com" }}]
 dependencies = [
-    "uipath>=2.10.0, <2.11.0"
+    "{_uipath_dependency_spec()}"
 ]
 requires-python = ">=3.11"
 """
@@ -41,7 +72,8 @@ requires-python = ">=3.11"
 
 def generate_uipath_json(target_directory):
     uipath_json_path = os.path.join(target_directory, UIPATH_CONFIG_FILE)
-    uipath_config = {"functions": {"main": "main.py:main"}}
+    project_id = resolve_existing_project_id(target_directory) or str(uuid.uuid4())
+    uipath_config = {"id": project_id, "functions": {"main": "main.py:main"}}
 
     with open(uipath_json_path, "w") as f:
         json.dump(uipath_config, f, indent=2)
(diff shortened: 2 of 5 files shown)


packages/uipath/src/uipath/_cli/cli_new.py
```diff
@@ -1,6 +1,9 @@
 import json
 import os
+import re
 import shutil
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _pkg_version
```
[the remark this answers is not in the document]
andreibalas-uipath: Checked against the repo's Ruff config: `ruff check` (with `force-single-line`-style isort settings in pyproject.toml) actually *requires* the two separate `from importlib.metadata import ...` lines and flags the combined form as I001. Kept the split form; `ruff check`, `ruff format --check`, and mypy are clean locally.

andreibalas-uipath:
## Manual end-to-end verification (local build 2.14.9, commit 85ea80c4; hint since reverted to `uipath init` in 141d107d)

Ran the UV-16117 repro against a wheel built from this branch, in a fresh venv, macOS / Python 3.13.

**1. Build & install**
```
$ uv build                       # → dist/uipath-2.14.9-py3-none-any.whl
$ uv venv && uv pip install dist/uipath-2.14.9-py3-none-any.whl
$ .venv/bin/uipath --version
uipath version 2.14.9
```

**2. `uipath new`**
```
$ uipath new demo
✓  Created 'main.py' file.
✓  Created 'pyproject.toml' file.
✓  Created 'uipath.json' file.
💡 Initialize project: uipath init          # (was `uv sync && uipath init` at 85ea80c4; reverted in 141d107d)
💡 Run project: uipath run main '{"message": "Hello World!"}'

$ grep uipath pyproject.toml
    "uipath>=2.14.0, <2.15.0"

$ cat uipath.json
{ "id": "3fa013dd-1d9e-4081-b88e-61c7f6a9ed03", "functions": { "main": "main.py:main" } }
```

**3. `uv sync`**
```
$ uv sync
 - uipath==2.14.9 (from file:///…/dist/uipath-2.14.9-py3-none-any.whl)
 + uipath==2.14.8
```
Stays on the 2.14 line — before this fix the same step printed `+ uipath==2.10.73`. The one-patch step to the released 2.14.8 happens only because 2.14.9 is not on PyPI yet; uv resolves from `pyproject.toml` and cannot see the locally installed wheel.

**4. `uipath init` / `uipath run` on the local 2.14.9 build**

To keep the local wheel through `uv sync`, a second copy of the scaffold got `[tool.uv.sources] uipath = { path = "…/uipath-2.14.9-py3-none-any.whl" }` appended; `uv sync` then reported `Checked 61 packages` with no change.
```
$ uipath --version
uipath version 2.14.9

$ uipath init
✓  Created '.env' file.
✓  Created 'bindings.json' file.
✓  Created 'entry-points.json' file with 1 entrypoint(s).
✓  Created 1 mermaid diagram file(s).
✓  Created 'project.uiproj' file.
✓  Created '.uipath/studio_metadata.json' file.
✓  Created AGENTS.md files and Claude Code skills.

$ uipath run main '{"message": "Hello World!"}'
output
└── message: Hello World!
✓  Successful execution.

$ uv pip show uipath | grep Version
Version: 2.14.9
```
No `'id' field not present in uipath.json` warning and no `id` backfill step — `new` wrote it. (`init`/`run` were also executed on the venv that `uv sync` had moved to the released 2.14.8; output was identical.)

**Baseline for comparison — released 2.14.8 (unfixed), same machine**
```
$ uipath --version
uipath version 2.14.8
$ uipath new baseline
💡  Initialize project: uipath init
'id' field not present in uipath.json
$ grep uipath pyproject.toml
    "uipath>=2.10.0, <2.11.0"
```

Acceptance criteria from UV-16117 all hold: pin derived from the running CLI, `uv sync` does not downgrade, `uipath.json` carries an `id`, hint has no stray space.

Note: the 4 failing `model-onboarding` checks are an upstream regression in `uipath-langchain-python` (`4b98dbaf`, fixed by UiPath/uipath-langchain-python#1050), unrelated to this change — the same jobs passed on every earlier SHA of this branch.

packages/uipath/docs/core/agents.md
```diff
@@ -119,7 +119,7 @@ version = "0.1.0"
 description = "..."
 authors = [{ name = "Your Name", email = "you@example.com" }]
 requires-python = ">=3.11"
-dependencies = ["uipath>=2.0", "uipath-langchain>=2.0"]
```
radu-mocanu, line 122: why was this needed?

packages/uipath/src/uipath/_cli/cli_new.py
```diff
@@ -79,9 +111,9 @@ def new(name: str):
         console.success(f"Created '{PYTHON_CONFIGURATION_FILE}' file.")
         generate_uipath_json(directory)
         console.success(f"Created '{UIPATH_CONFIG_FILE}' file.")
-        init_command = """uipath init"""
+        init_command = """uv sync && uipath init"""
```
radu-mocanu: uv sync should not be needed now

packages/uipath/docs/core/agents.md
andreibalas-uipath: Just a documentation inconsistency that was referencing an invalid `uipath-langchain` version.

radu-mocanu (approved):

PR #1872, opened 2026-08-26
feat(context-grounding): add start_deep_rag_from_attachments primitive

cfauchere:
## What

Adds a single-call primitive that creates the ephemeral index and starts the Deep RAG task in one server operation. Callers no longer need to create the index separately and wait for ingestion before starting the task.

## Changes

- **New SDK methods** `start_deep_rag_from_attachments(_async)` on `ContextGroundingService`, hitting `POST /ecs_/v2/deeprag/create` (vdbs `DeepRagController.CreateAsync` — the collection-level `Create` action registered in the OData model).
- **`CreateDeepRag` interrupt model** gains an optional `attachments: list[str] | None` field; the ephemeral-index validator now accepts `attachments` as an alternative to `index_id`.
- **`UiPathResumeTriggerCreator`** dispatches to the new SDK method when `CreateDeepRag.attachments` is set; the existing `start_deep_rag_ephemeral_async` / `start_deep_rag_async` paths are untouched.

## Why

The two-step flow (create ephemeral index → wait for ingestion → create Deep RAG task) collapses into one server call. On the client side that removes the middle `WaitEphemeralIndex` interrupt — the trailing wait for task completion remains.

## Notes

The endpoint URL is `POST /ecs_/v2/deeprag/create` based on vdbs' OData config (`EnableUnqualifiedOperationCall = true` + the collection action `Create` in `ODataModelConfiguration.cs:60`). Worth a smoke test against a running vdbs before merge to confirm the route resolves.


packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
```diff
@@ -1440,6 +1446,98 @@ async def start_deep_rag_ephemeral_async(
...
+        spec = self._deep_rag_from_attachments_creation_spec(
+            name=name,
+            prompt=prompt,
+            attachments=attachments,
+            citation_mode=citation_mode,
+            folder_key=folder_key,
+            folder_path=folder_path,
+        )
+
+        response = self.request(
```
CalebMartinUiPath, line 1484: what do we return if attachment isn't found? 404? should we catch that and return user readable message?

CalebMartinUiPath (approved):

radu-mocanu (approved):

packages/uipath-platform/src/uipath/platform/context_grounding/_context_grounding_service.py
cfauchere: the workflow just fails - maybe we validate attachment exist before we accept the request

cfauchere: Today the create endpoint returns 201 without touching Orchestrator — a bad attachment id only surfaces later as a Failed DeepRag status.

Fixing server-side in a follow-up vdbs PR: `OrchestratorClient.GetAttachmentByIdAsync` will translate the downstream 404 into `RecordNotFoundException` with a localized `Attachment_NotFound` message (mirrors the `SystemIndexesService.cs:82-85` pattern), and `CreateDeepRagFromAttachmentsAsync` will resolve the attachments up front so the failure is synchronous. That reaches the SDK as a real 404, which `BaseService.request` already wraps in `EnrichedException` (with the localized message pulled by `error_info`) — so no SDK change needed here.

PR #1875, opened 2026-08-27
feat(cli): add `uipath run --server-mode` for the pooled coded-agent lane

eduard-dumitru:
## What
Adds `uipath run --server-mode --ipc-pipe <name>` — keeps the `uipath run` process alive and serves subsequent jobs over uipath-ipc (no HTTP channel, no ready-ACK), reporting each job's real exit code. `uipath server` is left generic and unchanged.

## Why
The pooled server can't pre-warm generically — you don't know what to load until you have the package — so the natural model is `uipath run <job> --server-mode`: prime on the first run, then stay alive to serve more over IPC. Putting the mode on the `run` verb keeps `uipath server` pristine and makes it clear what's being run.

## Scope & safety
- Additive and default-off: the exit-code surfacing is threaded through a `surface_exit_code` parameter (default `False`), so `uipath server` and the Low-Code path are byte-for-byte unchanged.
- No new files; ~60 lines across `cli_run.py`, `cli_server.py`, `cli_server_ipc.py`, `_server_core.py`.

## Testing
- Unit tests for exit-code surfacing (on/off) + the `--server-mode` guard, and an end-to-end IPC test driving `start_ipc_server(surface_exit_code=True)` over a real named pipe.
- ruff / format / mypy clean; full suite green (2392 passed, 1 skipped).

### DIFF
--- a/packages/uipath/src/uipath/_cli/cli_run.py
+++ b/packages/uipath/src/uipath/_cli/cli_run.py
@@ -115,6 +115,18 @@ class _RunDiscoveryError(EntrypointDiscoveryException):
     default=None,
     help="Simulation config as a JSON object (same schema as simulation.json)",
 )
+@click.option(
+    "--server-mode",
+    is_flag=True,
+    help="Serve jobs over uipath-ipc and stay alive (requires --ipc-pipe), "
+    "instead of running once and exiting.",
+)
+@click.option(
+    "--ipc-pipe",
+    type=str,
+    default=None,
+    help="Named pipe for the uipath-ipc channel (used with --server-mode).",
+)
 @track_command("run")
 def run(
     entrypoint: str | None,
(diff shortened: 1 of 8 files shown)


packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -129,8 +141,21 @@ def run(
...
     simulation: str | None,
+    server_mode: bool,
+    ipc_pipe: str | None,
 ) -> None:
-    """Execute the project."""
+    """Execute the project, or serve jobs over uipath-ipc with --server-mode."""
+    if server_mode:
+        if not ipc_pipe:
+            console.error("--server-mode requires --ipc-pipe.")
+            return
```
radu-mocanu: nit: `console.error` already raises. no need for `return`

packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -129,8 +141,21 @@ def run(
...
+        # cli_server -> _server_core -> cli_run cycles at import time; defer it.
+        from .cli_server import preload_modules, run_ipc_server
+
+        preload_modules()
```
radu-mocanu: this will only load a few modules used by our SDK (a very small subset)
a coded agent can use ANY python library.
subsequent runs may (and probably will) still be very slow

we need to document how a developer can define a list of preloaded modules/generate them automatically at pack time

eduard-dumitru: as discussed separately, this is out of scope

PR #1876, opened 2026-08-31
fix(deps): allow uipath-runtime 0.13.x on the 2.13 line

radu-mocanu:
Releases `uipath` 2.13.23 from the 2.13 line with a single dependency-requirement change.

```diff
-version = "2.13.22"
+version = "2.13.23"
-  "uipath-runtime>=0.12.2, <0.13.0",
+  "uipath-runtime>=0.12.2, <0.14.0",
```

## Why

Every release on the 2.13 line caps `uipath-runtime` below 0.13.0, and 2.14.0 is the first release that widened it. A consumer that has to stay on the 2.13 line therefore cannot install `uipath-runtime` 0.13.x at all: there is exactly one `uipath-runtime` in an environment, and 0.13.2 does not satisfy `<0.13.0`, so resolution is unsatisfiable.

Published metadata is immutable, so the cap cannot be lifted by editing 2.13.22. It needs a new release on the line, which is why this exists. The source is byte-identical to 2.13.22.

## Branch

Cut from `c9b2428a`, the commit that built 2.13.22, per the open-source hotfix procedure. Base branch `base/uipath-2.13.x` is that same commit. Not for `main`, which is already on 2.14.x.

## Verification

- `uv lock` in `packages/uipath` pinned at `uipath-runtime==0.13.2`
- `uv sync --locked --all-extras` clean
- full `packages/uipath` suite green against `uipath-runtime` 0.13.2, exit 0, no failures

The lockfile is pinned at 0.13.2 rather than the newest 0.13.x so CI exercises the exact runtime this widening is being requested for.

## Publishing

CD is `workflow_dispatch`, so dispatch it against the hotfix branch after merge. 2.13.23 is free on PyPI, so this is a patch increment rather than a post-release.

### DIFF
--- a/packages/uipath/pyproject.toml
+++ b/packages/uipath/pyproject.toml
@@ -1,12 +1,12 @@
 [project]
 name = "uipath"
-version = "2.13.22"
+version = "2.13.23"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
 dependencies = [
   "uipath-core>=0.5.30, <0.6.0",
-  "uipath-runtime>=0.12.2, <0.13.0",
+  "uipath-runtime>=0.12.2, <0.14.0",
   "uipath-platform>=0.2.14, <0.3.0",
   "click>=8.3.1",
   "httpx>=0.28.1",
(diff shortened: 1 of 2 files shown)


radu-mocanu:
Closing: not needed, and not implementable under the single-minor range rule.

Making the 2.13 line accept `uipath-runtime` 0.13.x requires `>=0.13.0, <0.14.0`, not the two-minor `>=0.12.2, <0.14.0` proposed here. That drops 0.12 support outright, which is a minor-level change rather than a patch, and it is exactly what 2.14.0 already did in `94e43708`.

The consumer moves to 2.14.1 instead, which is the first published release admitting 0.13.x in a single-minor range. No release is needed on the 2.13 line.

PR #1880, opened 2026-09-01
feat(telemetry): emit lifecycle events for coded runs [PC-4888]

tudormatei1:
Coded agent and function runs emitted no telemetry once a job key was present, so a run in Orchestrator was invisible in App Insights. Observer hit exactly this when it tried to add coded agents and found nothing to query (AGVSOL-5022).

Three things change:

- Telemetry goes to the App Insights matching the environment you authenticated against. The connection string is picked from the host of `UIPATH_URL`, so an alpha login reports to alpha. An unauthenticated run, an Automation Suite deployment, or any unrecognised host falls back to production.
- `CodedAgentRun.*` and `CodedFunctionRun.*` now fire from the run path, Start plus a terminal event, so a cloud run is no longer silent. Low-code runs are skipped because uipath-agents-python already emits `AgentRun.*` for those.
- Every event carries the identity Observer asked for: folder, job, process, project and trace ids, so a run can be attributed without an Orchestrator round-trip. Coded agents and coded functions are told apart by the event name.

Three repo secrets need to exist before this merges, because `APPLICATIONINSIGHTS_CONNECTION_STRING` is no longer read by the build:

| Secret | Connection string from |
| --- | --- |
| `APPLICATIONINSIGHTS_CONNECTION_STRING_ALPHA` | `agents-alp-appins-ne-appins` |
| `APPLICATIONINSIGHTS_CONNECTION_STRING_STAGING` | `agents-stg-appins-ne-appins` |
| `APPLICATIONINSIGHTS_CONNECTION_STRING_PROD` | `agents-prd-appins-ne-appins` |

They are read by `build-package.yml` and `publish-dev.yml`, and passed through from `cd.yml`. A missing slot does not fail the build: the placeholder substitutes to an empty string and the wheel ships with that environment's telemetry switched off, so it goes quiet rather than reporting to the wrong instance. Worth adding all three in one go. The old secret can stay, since the integration tests still use it.

The new event names also need speccing in the BOI portal before they carry real volume, otherwise they land in the NotApplicable bucket. Failed events carry the error code, title and category but not the message or traceback, which for a coded run hold customer source and values. Say if you would rather match low-code exactly there.

Verified against staging: Start, End and Failed all observed in `agents-stg` with the full scope block, and a dashed `UIPATH_TRACE_ID` coming out normalized to hex so it joins the LLMOps trace.


packages/uipath/src/uipath/_cli/_run_telemetry.py
```diff
@@ -0,0 +1,222 @@
...
+SOURCE = "uipath-python-cli"
+
+NOT_AVAILABLE = "N/A"
+
+_LOWCODE_AGENT_TYPE = "uipath_lowcode"
+_FUNCTION_FRAMEWORK = "python"
+
+_ENV_IMAGE_VERSION = "IMAGE_VERSION"
+_ENV_AS_CLUSTER_ID = "AUTOMATION_SUITE_CLUSTER_ID"
+_ENV_AS_CLUSTER_VERSION = "AUTOMATION_SUITE_CLUSTER_VERSION"
```
radu-mocanu: let s be consistent with the env vars. please move them here since those are platform specific https://github.com/UiPath/uipath-python/blob/333ce46c60168aeadd8598563e73a615e545408c/packages/uipath-platform/src/uipath/platform/constants/__init__.py#L25

packages/uipath/src/uipath/_cli/_run_telemetry.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def _scope_properties() -> dict[str, Any]:
+    """Identity and scope, on every event so each one stands alone.
+
+    The cloud identity trio is required, so it reports ``N/A`` rather than being
+    dropped; optional scope is omitted, so a local run has no folder. Note
+    ``UIPATH_PROCESS_KEY`` holds a display name, and ``TraceId`` needs
+    :func:`resolve_trace_id` to normalize a dashed UUID into the hex spans use.
+    """
```
radu-mocanu, line 69: nit: keep docstring less verbose when possible

```suggestion
    """Identity and scope, on every event so each one stands alone.

    The cloud identity trio is required, so it reports ``N/A`` rather than being
    dropped; optional scope is omitted, so a local run has no folder.
    """
```

packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -221,6 +222,7 @@ async def execute() -> None:
                         chat_runtime: UiPathRuntimeProtocol | None = None
                         factory: UiPathRuntimeFactoryProtocol | None = None
                         governance_bootstrap: GovernanceBootstrap | None = None
+                        run_telemetry: RunTelemetry | None = None
```
radu-mocanu, line 225: `cli_debug` needs the same logic. this is the cli command used on `StudioWeb` runs

radu-mocanu (commented):
https://github.com/UiPath/uipath-python/pull/1880/changes#diff-937f88fc69a125f45f1631b8c4323919e8f592ad85e8eb50bea91c43cfabfb48R119

won t we have redundant telemetry events for local runs since we are emitting from this decorator as well? 
if that is the case we should be cost efficient and remove it from `cli_run` and `cli_debug` ` to avoid duplications

tudormatei1:
> https://github.com/UiPath/uipath-python/pull/1880/changes#diff-937f88fc69a125f45f1631b8c4323919e8f592ad85e8eb50bea91c43cfabfb48R119
> 
> won t we have redundant telemetry events for local runs since we are emitting from this decorator as well? if that is the case we should be cost efficient and remove it from `cli_run` and `cli_debug` ` to avoid duplications

we dont double emit as the Cli.Run contians less info and also Cli.RUN is not emitted on the cloud uipath run (cause the check with the job key is still present) https://github.com/UiPath/uipath-python/blob/c805809cd89f42dc199d713824757b07292c5f86/packages/uipath/src/uipath/_cli/_telemetry.py#L183. double emit only happens on local runs but that should be alrgiht?

packages/uipath/src/uipath/telemetry/_environment.py
```diff
@@ -0,0 +1,52 @@
...
+DEFAULT_ENVIRONMENT: UiPathEnvironment = "cloud"
+
+_ENVIRONMENT_BY_DOMAIN: dict[str, UiPathEnvironment] = {
+    "alpha.uipath.com": "alpha",
+    "staging.uipath.com": "staging",
+    "cloud.uipath.com": "cloud",
+}
+
+
+def _is_domain_or_subdomain(hostname: str, domain: str) -> bool:
```
radu-mocanu: subdomain is not a valid scenario. we can remove this

radu-mocanu:
> > https://github.com/UiPath/uipath-python/pull/1880/changes#diff-937f88fc69a125f45f1631b8c4323919e8f592ad85e8eb50bea91c43cfabfb48R119
> > won t we have redundant telemetry events for local runs since we are emitting from this decorator as well? if that is the case we should be cost efficient and remove it from `cli_run` and `cli_debug` ` to avoid duplications
> 
> we dont double emit as the Cli.Run contians less info and also Cli.RUN is not emitted on the cloud uipath run (cause the check with the job key is still present)
> 
> https://github.com/UiPath/uipath-python/blob/c805809cd89f42dc199d713824757b07292c5f86/packages/uipath/src/uipath/_cli/_telemetry.py#L183
> 
> . double emit only happens on local runs but that should be alrgiht?

we will have duplicate data either way. no reason to emit an additional event with less info

packages/uipath/src/uipath/_cli/_run_telemetry.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        status = result.status if result else UiPathRuntimeStatus.SUCCESSFUL
+
+        if status == UiPathRuntimeStatus.FAULTED:
+            contract = result.error if result else None
+            error_type = contract.code if contract else "Unknown"
+            self._emit(
+                "Failed",
+                {"Status": "Failed", **_error_properties(error_type, contract)},
+            )
```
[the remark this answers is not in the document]
radu-mocanu: this is a good find. so the problem is specifically that ErrorType means “error code” for a faulted result but “Python exception class” for a raised error. ErrorCode is already the consistent field for the code.

packages/uipath/src/uipath/_cli/_run_telemetry.py
```diff
@@ -0,0 +1,222 @@
...
+            contract = None
+        self._emit(
+            "Failed",
+            {
+                "Status": "Failed",
+                **_error_properties(type(exception).__name__, contract),
+            },
+        )
+
+    def _emit(self, suffix: str, extra: dict[str, Any]) -> None:
```
radu-mocanu:
```suggestion
    def _emit(self, suffix: str, extra: dict[str, Any],  include_duration: bool = False) -> None:
```

the current impl assumes every event that doesn t have the `start` suffix is a terminal one, thus it needs a duration. let's make that an opt-it. 
```
 if include_duration:
          properties["DurationMs"] = ...
```

packages/uipath/src/uipath/_cli/cli_run.py
tudormatei1: addressed.

packages/uipath/src/uipath/_cli/_run_telemetry.py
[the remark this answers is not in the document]
tudormatei1: addressed.

packages/uipath/src/uipath/_cli/_run_telemetry.py
tudormatei1: addressed.

packages/uipath/src/uipath/_cli/_run_telemetry.py
tudormatei1: addressed.

packages/uipath/src/uipath/_cli/_run_telemetry.py
tudormatei1: addressed.

packages/uipath/src/uipath/telemetry/_environment.py
tudormatei1: addressed.

radu-mocanu (approved):

PR #1883, opened 2026-09-07
feat(cli): stream job logs and result to the handler over uipath-ipc

eduard-dumitru:
## What

uipath-python owns the **uipath-ipc contract and connection** (`_job_api.py`) and points uipath-runtime's in-memory sinks (from #167) at the handler, so a Python serverless job streams its **logs and result** back over CoreIPC instead of the files. Covers **both** transport roles. Version `2.14.10 → 2.14.11`.

## Two roles, one contract — `IJobInvocationCommonApi` (`SendLog` + `SetResult`)

- **Pooled (`uipath server`)** — `Register(message)` grabs the handler's callback (`message.client.get_callback(IJobInvocationCommonApi)`) on the existing pooled pipe and installs the runtime sinks for the job. Gated per-job by a `StreamOutputOverIpc` flag on the run request; jobs run in-process (serialized), so the process-global sinks genuinely reach the callback.
- **Non-pooled (`uipath run --handler-ipc-pipe`)** — dials the handler's per-job server (`get_proxy`) and installs the sinks on a **dedicated loop/thread**, so the result's blocking ack can't deadlock the job's own event loop.

`SendLog` carries logs; `SetResult` carries a bounded envelope + a pointer to the output-arguments file the runtime spilled (the payload never crosses IPC). Best-effort and version-guarded: a no-op if the runtime lacks `output_sinks` or uipath-ipc is absent, so an older runtime keeps the files.

## Related PRs — one endeavor (merge in this order)

Three PRs, one endeavor — stream Python serverless job logs + result to the .NET Robot handler over CoreIPC:

1. **uipath-runtime** — the in-memory log/result sinks the host installs: UiPath/uipath-runtime-python#167
2. **uipath-python** — owns the uipath-ipc contract; installs those sinks over the pipe: UiPath/uipath-python#1883 ← this PR
3. **hdens** — the .NET handler side: UiPath/hdens#8031

## Testing

CLI/IPC tests green, incl. the non-pooled result-ack (deadlock) regression + DTO wire-key tests; ruff/format/mypy clean.

## Jira

Tracking: **ROBO-5981**.


packages/uipath/src/uipath/_cli/cli_server_ipc.py
```diff
@@ -57,7 +71,23 @@ async def StopJob(self, request: PythonServerStopJobRequest) -> bool:
...
+        self._callback: Any = None
+        self._loop: "asyncio.AbstractEventLoop | None" = None
+
+    async def Register(self, message: "Message[None]") -> bool:
+        client = message.client
+        if client is not None:
+            try:
+                from ._job_api import IJobInvocationCommonApi
+
+                self._callback = client.get_callback(IJobInvocationCommonApi)  # type: ignore[type-abstract]
```
robert-ursu: The callback is captured once on the singleton service, but `message.client` is the per-accept connection. hdens calls `Register` only from `Initialize()` and never re-issues it, so after a pipe reconnect (`_callback` proxies a closed connection) or a Python server restart (`_callback` is `None`) every opted-in job for the rest of the pod's life either streams into a dead connection or skips streaming. Since hdens no longer watches `execution.log` for these jobs, their logs vanish with no signal.

Suggest declaring `RunJob(self, request, message: Message)` and deriving `message.client.get_callback(...)` plus the running loop per request, instead of storing connection state on the service.

packages/uipath/src/uipath/_cli/cli_server_ipc.py
```diff
@@ -80,9 +110,27 @@ async def RunJob(self, request: PythonServerRunRequest) -> PythonServerRunJobRes
             f"Running job {_run_id(request.JobKey, request.ResumeVersion)}: {command_name} {args}"
         )
 
+        # Only when opted in and a callback exists. Install/clear the sinks INSIDE the job core's lock
+        # (via the hooks) so they're bound only while THIS job runs.
+        callback, loop = self._callback, self._loop
+        on_run_start: "Any" = None
+        on_run_end: "Any" = None
+        if request.StreamOutputOverIpc and callback is not None and loop is not None:
```
robert-ursu: When the handler opted the job in (it has already skipped the `execution.log` watcher) but no callback/loop is available here, the job silently runs with file sinks nobody reads. The mismatch is detectable at this point; it should surface as an error in the `RunJob` result (or at least a warning) rather than fall through.

packages/uipath/src/uipath/_cli/cli_server_ipc.py
```diff
@@ -108,11 +156,23 @@ async def start_ipc_server(pipe_name: str) -> None:
...
-    async with server:
-        await server.serve_forever()
+    try:
+        async with server:
+            await server.serve_forever()
+    finally:
+        # Drop the sinks so they can't outlive this loop (matters on in-process restart).
+        from ._job_api import clear_runtime_sinks
+
+        clear_runtime_sinks()
```
robert-ursu: This is a no-op: the sinks are `ContextVar`s set inside the per-request handler task (uipath-ipc dispatches each request via `asyncio.create_task`), so clearing them from the server task's context can't affect them. Same for `on_run_end=clear_runtime_sinks`. The "process-global sinks" framing in the comment and PR description is wrong; suggest dropping this and fixing the comment so the next reader doesn't rely on it.

packages/uipath/src/uipath/_cli/_job_api.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+def _drain(future: "Future[object]") -> None:
+    try:
+        future.exception()
+    except BaseException:
+        pass
+
+
+class _IpcLogHandler(logging.Handler):
```
robert-ursu, line 172: No self-exclusion filter. `uipath_ipc` is imported before the interceptor attaches this handler to every logger, so `uipath_ipc.client.connection`'s own diagnostics (`dropping malformed RESPONSE frame`, `receive loop failed`, ...) become `JobLogDto` entries in the customer's job output, and each one schedules another `SendLog` on the very transport whose failure it reports. Suggest a filter dropping `uipath_ipc.*` and this module's records (route them to the original stderr).

packages/uipath/src/uipath/_cli/_job_api.py
```diff
@@ -0,0 +1,289 @@
...
+        self._loop = loop
+
+    def emit(self, record: logging.LogRecord) -> None:
+        try:
+            message = self.format(record)
+            dto = JobLogDto(Message=message, LogLevel=_to_log_level(record.levelno))
+            future = asyncio.run_coroutine_threadsafe(
+                self._callback.SendLog(self._job_id, dto), self._loop
+            )
+            future.add_done_callback(_drain)
```
robert-ursu: The `SendLog` futures are never awaited, so the tail of the log can be dropped. Non-pooled: `disconnect_handler_ipc` -> `aclose()` fails any still-queued sends with `ConnectionError`, which `_drain` swallows. Pooled: the `RunJob` response is written right after `on_run_end` with no wait for acks, so .NET can `Unregister` the job before the last frames arrive (the sink then logs "Dropped pooled Python job log for unregistered job"). Track the futures on the handler and do a bounded wait before `aclose()` / before returning from `RunJob`.

packages/uipath/src/uipath/_cli/_job_api.py
```diff
@@ -0,0 +1,289 @@
...
+    handler = _IpcLogHandler(job_id, callback, loop)
+    handler.setFormatter(logging.Formatter("%(message)s"))
+
+    def _result_sink(result: Any, output_arguments_file_path: str) -> None:
+        dto = _to_result_dto(job_id, result, output_arguments_file_path)
+        try:
+            future = asyncio.run_coroutine_threadsafe(
+                callback.SetResult(job_id, dto), loop
+            )
+            future.result(timeout=_SET_RESULT_TIMEOUT_S)
```
robert-ursu: Two issues with result delivery:

1. The `bool` from `SetResult` is discarded. hdens returns `false` when `registry.TryGet(jobId)` fails or `executor.SetResult` throws, so a rejected result is treated as delivered.
2. The `logger.exception` on line 189 can only reach `_IpcLogHandler` at this point (the interceptor has removed every other handler and redirected stderr), so the "must not be swallowed" notice rides the channel that just failed and is dropped. Write it to the interceptor's original stream (`sys.__stderr__`) or into the result envelope.

packages/uipath/src/uipath/_cli/_job_api.py
```diff
@@ -0,0 +1,289 @@
...
+        """Close the client and stop its loop/thread."""
+        try:
+            asyncio.run_coroutine_threadsafe(self._client.aclose(), self._loop).result(
+                timeout=_SET_RESULT_TIMEOUT_S
+            )
+        except Exception:
+            pass
+        self._loop.call_soon_threadsafe(self._loop.stop)
+        self._thread.join(timeout=_SET_RESULT_TIMEOUT_S)
+        self._loop.close()
```
robert-ursu: If `join(timeout=30)` times out because the IPC thread is wedged (peer half-open, `aclose()` hanging), `loop.close()` raises `RuntimeError: Cannot close a running event loop`. Inside the context manager's `finally` that replaces the job's real exception, or turns a successful run into exit 1. Same issue on the failure path in `connect_handler_ipc` (line 266), where the bare `raise` never executes and the original `TimeoutError` is masked. Guard `close()` on `thread.is_alive()` and make teardown non-raising.

packages/uipath/src/uipath/_cli/_job_api.py
```diff
@@ -0,0 +1,289 @@
...
+        ) from e
+
+    loop = _new_ipc_event_loop()
+    thread = threading.Thread(
+        target=loop.run_forever, name="uipath-handler-ipc", daemon=True
+    )
+    thread.start()
+
+    async def _build() -> Any:
+        client = IpcClient(transport=NamedPipeClientTransport(pipe))
```
robert-ursu: `NamedPipeClientTransport` connects lazily on the first call, so nothing here verifies the pipe is reachable. But installing the handler makes `UiPathRuntimeLogsInterceptor` skip the `execution.log` file handler, so an unreachable pipe (POSIX `$TMPDIR` mismatch, peer not yet bound, typo) loses 100% of the job's logs while `uipath run` exits 0 with "Successful execution". Do one eager round-trip at connect and fail loudly, or fall back to the file handler on first failure.

Also: no `request_timeout` on `IpcClient`. Against a peer that accepted the socket but stalls, every log line becomes an immortal task parked on the write lock with no cap, for the whole job.

packages/uipath/pyproject.toml
```diff
@@ -1,12 +1,12 @@
 [project]
 name = "uipath"
-version = "2.14.10"
+version = "2.14.11"
```
robert-ursu: 2.14.11 is already on PyPI (main is at 2.14.14), and the branch conflicts with main on `pyproject.toml` and `uv.lock`. Needs a rebase and a bump to >= 2.14.15. Worth getting right because the hdens `VenvIpcCapability` version floors will be set from this release: if they end up at 2.14.11, pods with 2.14.11-2.14.14 venvs would get `StreamOutputOverIpc` / `--handler-ipc-pipe`, the watcher skipped, and an older `uipath` that ignores or rejects the option.

packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -212,8 +219,14 @@ async def execute() -> None:
                         JsonLinesFileExporter(ctx.trace_file)
                     )
 
-                async with ResourceOverwritesContext(
-                    lambda: read_resource_overwrites_from_file(ctx.runtime_dir)
+                # If a pipe was given, install the sinks around the run (always torn down); else a no-op.
+                from ._job_api import handler_ipc_connection
+
+                async with (
+                    handler_ipc_connection(handler_ipc_pipe, ctx.job_id or ""),
```
robert-ursu: `ctx.job_id or ""` sends an empty string where the .NET contract has `Guid jobId`. Without `UIPATH_JOB_KEY` (manual/dev runs) Newtonsoft can't convert `""` to `Guid`, so every `SendLog`/`SetResult` errors and, given the issues above, nothing is reported. Skip connecting, or fail fast with a clear message, when `ctx.job_id` is `None`.

packages/uipath/src/uipath/_cli/cli_run.py
```diff
@@ -129,6 +135,7 @@ def run(
     debug_port: int,
     keep_state_file: bool,
     simulation: str | None,
+    handler_ipc_pipe: str | None,
```
robert-ursu, line 140: `handler_ipc_pipe` isn't forwarded to `Middlewares.next("run", ...)`, and the sinks are installed only inside `execute()`. A `run` middleware that returns `should_continue=False` executes the job with the flag silently ignored, while the handler has already skipped the `execution.log` watcher. Either pass it through the middleware chain or refuse the flag when a run middleware takes over.

robert-ursu (commented):
Reviewed against the PR head, the uipath-runtime 0.13.5 / uipath-ipc 2.5.2 sources, and the hdens#8031 C# side.

Main theme: the new IPC path removes the `execution.log` fallback but has no way to notice when the channel is broken, so logs can vanish while the job still reports success. The two blocking items are the stale `Register` callback on the pooled server and the lazy connect on the non-pooled path. Details inline.

robert-ursu (approved):

PR #1886, opened 2026-09-07
fix(cli): add --type to uipath new so function projects stay reachable

### LINKED ISSUE #1543: feat(cli): add --type flag to `uipath new` to support function project scaffolding alongside agents
## Problem

When `uipath-langchain` is installed alongside `uipath`, running `uipath new <name>` always creates a **LangGraph agent project** regardless of what the user intended. This happens because `langgraph_new_middleware` in `uipath-langchain` intercepts the `new` command unconditionally and returns `should_continue=False`, completely bypassing the base function scaffolding logic.

```python
# uipath_langchain/_cli/cli_new.py
def langgraph_new_middleware(name: str) -> MiddlewareResult:
    # always runs, no way to opt out
    with console.spinner(f"Creating new agent {name} in current directory ..."):
        generate_script(directory)   # creates langgraph.json + agent main.py
        generate_pyproject(directory, name)  # uipath-langchain dep
    return MiddlewareResult(should_continue=False)  # base logic never runs
```

There is no way today to scaffold a pure Python **function** project (the `EchoIn`/`EchoOut` dataclass pattern from the base template) when `uipath-langchain` is installed.

## Context

The `uip` CLI (UiPath/cli) has a `functions-tool` plugin that routes `uip functions new --language py` to `uipath new <name>`. Since `functions-tool` is explicitly for **Functions** (not agents), it always expects a function project to be scaffolded. With `uipath-langchain` installed the passthrough silently creates an agent project instead.

## Proposed solution

**1. Add `--type` option to `uipath new` in the base package (`packages/uipath/src/uipath/_cli/cli_new.py`):**

```python
@click.command()
@click.argument("name", type=str, default="")
@click.option(
    "--type",
    "project_type",
    type=click.Choice(["function", "agent"]),
    default="function",
    help="Project type to scaffold: function (default) or agent.",
)
@track_command("new")
def new(name: str, project_type: str):
    ...
    result = Middlewares.next("new", name, project_type=project_type)
```

**2. Update `langgraph_new_middleware` in `uipath-langchain` to respect the flag:**

```python
def langgraph_new_middleware(name: str, project_type: str = "agent") -> MiddlewareResult:
    if project_type == "function":
        return MiddlewareResult(should_continue=True)  # pass through to base logic
    # existing agent scaffolding unchanged ...
```

## Expected behaviour after fix

| Command | Result |
|---|---|
| `uipath new my-fn` | function project (base template — default) |
| `uipath new my-fn --type function` | function project (explicit) |
| `uipath new my-agent --type agent` | LangGraph agent project (existing behaviour) |

## Why default to `function`

The base `uipath` package is language-agnostic and its own `new` command already creates a function project (the `EchoIn`/`EchoOut` template). The agent scaffolding lives in `uipath-langchain` as an opt-in extension. Defaulting to `function` preserves the base package's intent; users who want an agent project pass `--type agent` (or install `uipath-langchain` and rely on a future `uipath new --type agent` alias).

## Files to change

- `packages/uipath/src/uipath/_cli/cli_new.py` — add `--type` Click option, pass to middleware
- `uipath-langchain-python/src/uipath_langchain/_cli/cli_new.py` — `langgraph_new_middleware` accepts and respects `project_type`

vldcmp-uipath:
## Problem

When an agent framework integration such as `uipath-langchain` is installed, its entry-point middleware claims `uipath new` unconditionally (`should_continue=False` in all cases), so the base function scaffold is unreachable — `uipath new <name>` always produces an agent project with no `uipath.json`. Fixes #1543.

## Changes

The fix lives in the dispatch, not in the integrations. The `new` middleware chain is the only route to an agent scaffold, and [cli_new.py](https://github.com/UiPath/uipath-python/blob/main/packages/uipath/src/uipath/_cli/cli_new.py) is its only caller, so the base CLI decides whether the frameworks are consulted at all:

- **`--type auto`** (the new default) preserves today's behaviour — an installed framework claims the scaffold; with none installed, the base function project is created.
- **`--type function`** never calls the chain, so no installed framework can intercept it. This is the option callers needing a guaranteed function project (e.g. the `uip` functions-tool passthrough) should pass.
- **`--type agent`** calls the chain and errors when nothing claims it, instead of quietly producing a function project. It is a guarantee, which is what makes it useful in CI.

- **`--agent-framework`** picks between installed frameworks: `uipath new my-agent --type agent --agent-framework uipath-langchain`. It takes the integration package, spelled exactly as the error messages list it, is only valid with `--type agent`, and dispatches to that framework alone so another cannot claim the scaffold first. With one framework installed there is nothing to choose and it is optional; with several and no choice given, the command warns — naming the installed frameworks and the one it picked — and scaffolds with the first discovered, preserving the pre-`--type` behaviour rather than breaking those environments.

**There is still no list of frameworks in the base package** — no enum, no framework-to-package map, nothing to update when an eighth integration ships. The frameworks that exist, and the names `--agent-framework` accepts, are derived from the registered `new` middlewares and the distributions that registered them, so a framework this CLI has never heard of is named and selected like any other.

**Integrations need no changes.** `Middlewares.next("new", name)` is byte-identical to today's call, so existing releases keep working as they are, with no new `uipath` floor, no kwargs, and no legacy-signature warnings.

## Behaviour

| Installed integrations | `uipath new x` (auto) | `uipath new x --type function` | `uipath new x --type agent` |
|---|---|---|---|
| none | function project | function project | error: install an agent framework |
| one | that framework's agent | function project | that framework's agent |
| several | warns, scaffolds with the first discovered | function project | warns, scaffolds with the first discovered |

## Testing

- 21/21 in `tests/cli/test_new.py`, 1490/1490 across `tests/cli`; mypy and ruff clean.
- Verified end-to-end in a venv against **unmodified** `uipath-langchain` 0.17.8 from its `main`: `auto` scaffolds the LangGraph agent, `--type function` now reaches the base scaffold (the bug), `--type agent` scaffolds the agent. With a second integration installed, both `auto` and `agent` warn and scaffold with the first discovered framework while `--type function` still succeeds; `--agent-framework uipath-langchain` selects explicitly, and a value that is not an installed package lists the ones that are.
- Docs build clean; the rendered CLI reference shows the three `--type` choices.

### uipath

```toml
[project]
dependencies = [
  "uipath==2.14.23.dev1018867682",

  # "uipath>=2.14.23.dev1018860000,<2.14.23.dev1018870000",
]

[[tool.uv.index]]
name = "testpypi"
explicit = true

[tool.uv.sources]
```


packages/uipath/src/uipath/_cli/cli_new.py
```diff
@@ -57,10 +61,87 @@ def generate_uipath_json(target_directory):
...
+        return _select_agent_framework(installed, agent_framework).scaffold(name)
+
+    if len(installed) > 1:
+        console.error(
+            "Multiple agent frameworks are installed: "
+            + ", ".join(framework.package for framework in installed)
+            + f".\nPick one with `uipath new {name} --type agent "
+            "--agent-framework <framework>`, or run "
+            f"`uipath new {name} --type function` to create a function project."
+        )
```
radu-mocanu: this is a breaking change. let s warn that there are multiple frameworks installed but the scaffold should not fail

radu-mocanu (approved):

PR #1890, opened 2026-09-09
feat(cli): add an optional per-job scope to the server job core

viswa-uipath:
## What

The pre-warmed server runs every job through `_run_command_isolated`, which applies the job's env and cwd, runs the command, then restores the baseline. Nothing outside that module could bracket the window in which a job actually runs.

This adds `register_job_scope_provider(provider)`: an optional async context manager entered inside the per-job lock, **after** env/cwd are applied and **before** the command runs, and exited before that state is restored — so the scope observes the same environment and cwd the job itself sees.

With no provider registered (the default) the job path is byte-for-byte unchanged.

## Contract

- Best-effort: a provider that raises on entry means the job runs unscoped; one that raises on exit cannot change the job's result. Both are logged.
- Registered once at server startup; passing `None` clears it.
- Applies to both transports, since they share `_run_command_isolated` (`cli_server.py`, `cli_server_ipc.py`).

## Tests

Eight new cases in `tests/cli/test_server_job_core.py`: no-provider path, env/cwd visibility inside the scope, teardown ordering against the env restore, enter → command → exit ordering, exit on the command's failure path, both fail-open paths, and deregistration.

## Known limitations (follow-ups, not addressed here)

- `asyncio.to_thread` is not cancellable, so a cancelled job unwinds the scope (and the env/cwd restore) while the job thread is still running. Pre-existing for the env restore; the scope inherits it.
- `__aexit__` is always called with `(None, None, None)`, so a provider cannot observe that the job failed or was cancelled.
- The provider is entered/exited on the event loop thread while the command runs in a worker thread: contextvars propagate in, `threading.local` does not.


packages/uipath/src/uipath/_cli/_server_core.py
```diff
@@ -33,6 +36,66 @@ def init(self) -> None:
 
 _state = _ServerState()
 
+logger = logging.getLogger(__name__)
+
+# Optional per-job scope, registered by an out-of-process concern that needs to bracket
+# every job this server runs. ``None`` — the default — means no scope is installed and
+# the job runs exactly as it did before.
+_job_scope_provider: Callable[[], AbstractAsyncContextManager[None]] | None = None
```
radu-mocanu: can we put this on `_ServerState` instead? it s already initialized at startup, which is when the provider gets registered anyway. a module global can be flipped by any import while jobs are in flight.

packages/uipath/src/uipath/_cli/_server_core.py
```diff
@@ -33,6 +36,66 @@ def init(self) -> None:
...
+        scope = None
+
+    try:
+        yield
+    finally:
+        if scope is not None:
+            try:
+                # Shielded: a cancel delivered while the job unwinds must not leave the
+                # provider half torn down.
+                await asyncio.shield(scope.__aexit__(None, None, None))
```
radu-mocanu, line 99: shield re-raises CancelledError right away and leaves the teardown running detached, so the `finally` below restores env/cwd while it s still in flight. that breaks the "exits before the env is restored" guarantee, and since CancelledError isn t an Exception a failure in there is never logged.

```python
task = asyncio.ensure_future(scope.__aexit__(None, None, None))
try:
    await asyncio.shield(task)
except asyncio.CancelledError:
    await task
    raise
except Exception:
    logger.warning("job scope provider failed to exit", exc_info=True)
```

the cancellation test only checks the teardown finishes, not what env it saw, so it passes either way.

viswa-uipath: Wont a stuck `await task` block everything?

packages/uipath/src/uipath/_cli/_server_core.py
viswa-uipath: Done, looks a bit cleaner. Even _ServerState is module global only, it wont make a difference in behavior.

packages/uipath/src/uipath/_cli/_server_core.py
radu-mocanu: agreed, let s protect against that with a timeout then
`await asyncio.wait_for(asyncio.shield(task), timeout=5)`

we can do it in a subsequent PR

radu-mocanu (approved):

packages/uipath/src/uipath/_cli/_server_core.py
```diff
@@ -22,6 +25,13 @@ class _ServerState:
     def __init__(self) -> None:
         self.lock: asyncio.Lock | None = None
         self.baseline_env: dict[str, str] | None = None
+        # Optional per-job scope, registered by an out-of-process concern that needs to
+        # bracket every job this server runs. ``None`` — the default — means no scope is
+        # installed and the job runs exactly as it did before. Set here rather than in
+        # ``init`` so a provider registered before server startup is not cleared.
```
radu-mocanu: nit: please shorten this comment; it describes the implementation process

PR #1892, opened 2026-09-10
chore: normalize CRLF line endings in 5 testcase/script files

TeodorEscuUiPath:
## Summary
- Renormalizes line endings (CRLF -> LF) for 5 files that were committed with raw CRLF baked into the blob in `120384a2` (the uipath-core/uipath-platform monorepo consolidation).
- With `core.autocrlf=true`, Git's status/diff assumes LF-normalized storage; these files' stored CRLF desynced from that assumption, so `git status` reported them as modified on every fresh checkout even though the checked-out bytes were byte-for-byte identical to `HEAD`.
- Verified content is unchanged: diffing old vs. new blob with line endings stripped shows `IDENTICAL` for all 5 files. Pure whitespace/line-ending normalization, no code or content change.

Files:
- `packages/uipath/scripts/debug_test.py`
- `packages/uipath/testcases/apicalls-testcase/src/assert.py`
- `packages/uipath/testcases/basic-testcase/src/assert.py`
- `packages/uipath/testcases/common/trace_assert.py`
- `packages/uipath/testcases/langchain-cross/expected_traces.json`

Tagging @radu-mocanu since these files trace back to their commit (`120384a2`) for a quick sanity check.

## Test plan
- [x] `git status` clean on a fresh checkout of this branch (no spurious CRLF diffs)
- [x] Content verified byte-identical to previous commit once line endings are stripped (no code change)


radu-mocanu:
Let's also add this in `gitattributes`:

```gitattributes
* text=auto

*.py   text eol=lf
*.json text eol=lf
*.yaml text eol=lf
*.yml  text eol=lf
*.toml text eol=lf
```

Then we could run `git add --renormalize .` once to clean up any files that were already committed with CRLF.

 This PR can be the one-time normalization for the files that already slipped through.

radu-mocanu (approved):

PR #1894, opened 2026-09-10
fix(evals): include agent output in AgentRunHistory [AE-2147]

rakesh-uipath:
Fixes [AE-2147](https://uipath.atlassian.net/browse/AE-2147) — reported by James Dickson in [#help-agent-builder](https://uipath-product.slack.com/archives/C07SE86UZ6Y/p1787771136805009), filed by Maria Vimer.

## The bug

`{{AgentRunHistory}}` reaches the trajectory judge as a tool-call transcript with the agent's own output missing, so the judge grades an incomplete trace.

It is real, and it is one predicate:

https://github.com/UiPath/uipath-python/blob/30f6bec2/packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py#L653

```python
for span in workload_trace:
    if span.attributes and (tool_name := span.attributes.get(TOOL_NAME_ATTR)):
```

A span without `tool.name` is skipped, and the docstring said so out loud: *"Only processes tool spans"*. The workload's own answer lives on `WorkloadExecution.workload_output` (populated in `runtime.py:1143`) and never entered the string.

There is no second channel for it. On `BaseLLMTrajectoryEvaluator` the actual-output placeholder **is** the history:

```python
actual_output_placeholder: str = "{{AgentRunHistory}}"   # llm_judge_trajectory_evaluator.py:63
...
return trace_to_str(workload_execution.workload_trace)   # line 83 — output not passed
```

So the judge never saw the output at all — while its own system prompt promises it *"a trace/history of the agent's actions and outputs"*.

Reproduced against James's run (one `Web_Search` tool span, agent answered correctly). Before:

```
[2026-08-26 18:46:40] LLM Response:
  Agent Selected 1 Tool(s):

  Tool: Web_Search
  Arguments: {}

[2026-08-26 18:46:47] Tool Call Response - Web_Search:
{}
```

Byte-for-byte what his screenshot shows. The judge scored it 10% with the justification *"It never produced a user-facing answer identifying Argentina as the most recent World Cup winner"* — while the Actual Output panel on the same screen read `{"search_results_answer": "Argentina won the most recent FIFA World Cup (Qatar 2022)..."}`. The agent was right and got marked down for the serialiser's omission.

## The fix

`trace_to_str` takes the workload output and appends it as a final block; both trajectory evaluators pass it.

```
[2026-08-26 18:46:47] Tool Call Response - Web_Search:
{}

Agent Output:
{"search_results_answer": "Argentina won the most recent FIFA World Cup (Qatar 2022), ..."}
```

Fixed in `LLMJudgeTrajectoryEvaluator` / `LLMJudgeTrajectorySimulationEvaluator` and in `LegacyTrajectoryEvaluator`, which had the same defect through the same helper.

## Deliberately not changed

Intermediate assistant messages are still not rendered. LLM spans carry the full system prompt, and `test_legacy_trajectory_prompt_uses_compact_tool_history` exists precisely to keep a 10k-char prompt out of the judge's context. Surfacing intermediate reasoning needs a bounded projection of the LLM spans, not a filter removal — worth doing, separate change. James's other asks in that thread (documenting what each `{{...}}` binds to in the evaluator UI, eval-driven judge authoring) are product work, not this.

`workload_output=None` keeps the old behaviour, so no existing caller changes shape.

## Tests

`packages/uipath/tests/evaluators/test_trajectory_agent_output.py` — reproduces the AE-2147 run through both evaluators and asserts the answer reaches the prompt. Both fail on `main`:

```
E  AssertionError: assert 'Argentina won the most recent FIFA World Cup (Qatar 2022).' in
   '...Tool Call Response - Web_Search:\n{}\n'
```

Plus five cases on `trace_to_str` itself in `test_evaluator_helpers.py` (ordering, str vs dict output, omitted output, empty output).

## Verification

Ran locally, all from `packages/uipath`:

- `uv run pytest` — green, exit 0
- `uv run ruff check .` — All checks passed
- `uv run ruff format --check .` — 335 files already formatted
- `uv run mypy --config-file pyproject.toml .` — no issues in 335 source files
- `uv run python scripts/lint_httpx_client.py` — clean

Version bumped 2.14.14 → 2.14.15; 2.14.14 is already on PyPI so `check-version-availability` would have failed otherwise.

[AE-2147]: https://uipath.atlassian.net/browse/AE-2147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/uipath/pyproject.toml
+++ b/packages/uipath/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "uipath"
-version = "2.14.14"
+version = "2.14.15"
 description = "Python SDK and CLI for UiPath Platform, enabling programmatic interaction with automation services, process management, and deployment tools."
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11"
--- a/packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
+++ b/packages/uipath/src/uipath/eval/_helpers/evaluators_helpers.py
@@ -634,14 +634,34 @@ def tool_calls_output_score(
     ), justifications
 
 
-def trace_to_str(workload_trace: Sequence[ReadableSpan]) -> str:
+def _format_workload_output(workload_output: dict[str, Any] | str) -> str:
+    """Render a workload's own output for the run-history string."""
+    if isinstance(workload_output, str):
+        return workload_output
+    try:
+        return json.dumps(workload_output)
+    except (TypeError, ValueError):
+        return str(workload_output)
+
+
+def trace_to_str(
+    workload_trace: Sequence[ReadableSpan],
+    workload_output: dict[str, Any] | str | None = None,
+) -> str:
     """Convert OTEL spans to a platform-style workload run history string.
 
     Creates a similar structure to LangChain message processing but using OTEL spans.
-    Only processes tool spans (spans with 'tool.name' attribute).
+    Tool spans (spans with 'tool.name' attribute) are rendered as the trajectory;
+    the workload's own output is appended as a final "Agent Output" block.
+
+    LLM spans are deliberately not rendered — they carry the full system prompt
+    and would swamp the judge's context. `workload_output` is the workload's
+    answer, so it is passed in rather than recovered from the trace.
 
     Args:
         workload_trace: List of ReadableSpan objects from the workload execution
+        workload_output: The workload's own output. Omit it to render the tool
+            calls alone.
 
     Returns:
         String representation of the workload run history in platform format
@@ -704,5 +724,10 @@ def trace_to_str(workload_trace: Sequence[ReadableSpan]) -> str:
             )
             platform_history.append(f"{tool_result}")
             platform_history.append("")
+
+    if workload_output is not None:
+        platform_history.append("Agent Output:")
+        platform_history.append(_format_workload_output(workload_output))
+        platform_history.append("")
 
     return "\n".join(platform_history)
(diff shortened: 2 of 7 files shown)


Chibionos (commented):
## /deep-review — PR #1894 · fix(evals): include agent output in AgentRunHistory [AE-2147]

**Branch:** `fix/ae-2147-trajectory-agent-output` → `main` | **Size:** +309/−12 across 7 files

### Verdict
Approve — correct, minimal, well-tested bugfix; both lenses (architecture + runtime) clean, Python domain pillar clean, `pytest`/`ruff check`/`ruff format --check`/`mypy` all pass on the changed files.

### Slop meter
## 🎚️ Slop Meter — PR #1894

**Verdict:** 🟢 clean
**Slop axis:** 2/100 · **Tunnel-vision axis:** 0/100
**Diff:** +309 LOC across 7 files

Only signal: `hardcoded_version` (2 hits, `pyproject.toml` + `uv.lock`) — the legitimate version bump described in the PR body (2.14.14 already published to PyPI). No other slop or tunnel-vision signals fired.

### Codex recheck
Codex CLI is on PATH but the exec call failed with "You hit your spend cap set by the owner of your workspace" (not the documented "not on PATH" case). Treated as unavailable for this run — proceeding on local two-lens findings only (Codex not available this run — stronger guarantee once budget/config is fixed).

### Findings index (all inline below)

|---|-----|-----------|----------|
| — | — | — | No P0/P1/P2 findings |

(No finding bodies in the summary — every body would be an inline comment anchored to a line; there are none to anchor this cycle.)

PR #1895, opened 2026-09-12
feat: allow guardrail evaluation to carry attachment references

apetraru-uipath:
## What

Adds an optional `attachments` keyword to `GuardrailsService.evaluate_guardrail` and a `GuardrailAttachment` model (`id`, `fileName`, `mimeType`) so a caller can tell the guardrails backend which Orchestrator attachments a guardrail should inspect. When attachments are present the call also carries the run's folder key in `x-uipath-folderkey` (from the SDK config) so the backend can resolve them folder-scoped, and uses a 60 s timeout because the backend fetches and decodes each file inside the request.

Backward compatible: without `attachments` the request body is byte-identical to today. Bumps `uipath-platform` to 0.2.31 and `uipath` to 2.14.22 (its `SDK_REFERENCE.md` records the new parameter and its platform floor moves to 0.2.31).

## v2 — references instead of urls (commit `db3cd201`)

The first version carried a resolved SAS `url` per attachment and redacted it from the `@traced` span. The backend now resolves attachment ids itself through Orchestrator (see UiPath/Agents#6256), so the `url` field and the redaction processor are gone and the model is just `id` / `fileName` / `mimeType`. New: the folder-key header. 36 tests pass; ruff, ruff format and mypy clean.

## Review fixes

- **`e831652c` — SAS urls no longer reach the trace span.** `@traced` records a function's arguments on the OpenTelemetry span by default, so every `GuardrailAttachment.url` (a short-lived SAS credential) was landing in `input.value`. An `input_processor` now redacts `attachments[*].url` and leaves the rest of the span intact — tighter than `hide_input=True`. Thanks Copilot.
- `49b50329` — lockfiles relocked after the version bump (`uv lock --check` in CI).
- Added the two tests Copilot noted were missing: an attachment-bearing evaluation forwards the 60 s timeout; the default path does not.

## Notes for reviewers

- **Backward compatible.** Without `attachments`, or with an empty list, the request body is byte-identical to today.
- **Timeout.** A validate call carrying attachments waits for the backend to fetch and decode each file. The default client timeout is 30 s and `RequestSpec.timeout` was constructed but never forwarded, so this passes 60 s explicitly when attachments are present.

## Test plan

`packages/uipath-platform` — **35 tests pass** (+8 over baseline). `ruff check`, `ruff format --check`, `mypy src tests` clean.

Pairs with UiPath/Agents#6256.


packages/uipath-platform/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "uipath-platform"
-version = "0.2.28"
+version = "0.2.29"
```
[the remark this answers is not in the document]
apetraru-uipath: Fixed in 49b50329 — all three lockfiles relocked; CI's `uv lock --check` is green.

packages/uipath-platform/src/uipath/platform/guardrails/_guardrails_service.py
```diff
@@ -102,12 +110,18 @@ def evaluate_guardrail(
         self,
         input_data: str | dict[str, Any],
         guardrail: BuiltInValidatorGuardrail,
+        *,
+        attachments: list[GuardrailAttachment] | None = None,
```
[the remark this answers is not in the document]
apetraru-uipath: Fixed in e831652c: an `input_processor` redacts `attachments[*].url` before the span records inputs, leaving payload, guardrail and attachment identity intact. Two tests cover it, including a no-mutation check.

packages/uipath-platform/src/uipath/platform/guardrails/_guardrails_service.py
```diff
@@ -147,13 +163,22 @@ def evaluate_guardrail(
...
         }
+        # The default client timeout is 30s (common/_http_config.py). A validate call
+        # carrying attachments waits for the backend to fetch and decode each one, so give
+        # it more room. RequestSpec.timeout exists but is never forwarded, so pass it here.
+        request_kwargs: dict[str, Any] = {
+            "json": spec.json,
+            "headers": request_headers,
+        }
+        if attachments:
+            request_kwargs["timeout"] = _ATTACHMENT_VALIDATE_TIMEOUT_SECONDS
```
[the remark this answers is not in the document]
apetraru-uipath: Added in e831652c: one test asserts the attachment-bearing call forwards `timeout=60.0` (via `httpx_mock` request extensions), and one asserts the default path does not.

packages/uipath-platform/src/uipath/platform/guardrails/guardrails.py
```diff
@@ -96,6 +96,33 @@ class BuiltInValidatorGuardrail(BaseGuardrail):
     model_config = ConfigDict(populate_by_name=True, extra="allow")
 
 
+class GuardrailAttachment(BaseModel):
+    """A reference to a file attached to the run that a guardrail may inspect.
+
+    Passed to [`GuardrailsService.evaluate_guardrail`][uipath.platform.guardrails.GuardrailsService.evaluate_guardrail]
+    so the guardrails backend can read the file's contents rather than only its metadata.
+
+    Only the Orchestrator attachment id crosses the wire: helix resolves it through
```
valentinabojan, line 105: Don't refer the helix project name in the SDK docs please

packages/uipath-platform/src/uipath/platform/guardrails/_guardrails_service.py
```diff
@@ -141,19 +158,35 @@ def evaluate_guardrail(
         execution_source = self._execution_context.execution_source
         if execution_source:
             source_headers[HEADER_GUARDRAILS_SOURCE] = execution_source
+        # When attachments are present, tell helix which folder the run executed in
```
valentinabojan, line 161: Same here. Drop the helix terminology. In fact, do we need this comment at all? Let's reduce the comments' verbosity and keep only what's really needed and useful.

valentinabojan (approved):

packages/uipath-platform/src/uipath/platform/guardrails/guardrails.py
apetraru-uipath: Addressed in #1903 — project name dropped and the surrounding comments trimmed.

packages/uipath-platform/src/uipath/platform/guardrails/_guardrails_service.py
apetraru-uipath: Addressed in #1903 — project name dropped and the surrounding comments trimmed.

PR #1896, opened 2026-09-14
fix(eval): reject non-object simulated inputs and unwrap stringified objects (SRE-655743)

mjnovice:
## Problem

Synthetic input generation (`inputMockingStrategy`) could return a JSON **string** instead of the object the agent input schema asks for. Two paths in `mocks/_structured_output.py` allow it:

1. `ResponseFormatStructuredOutput.generate` → `json.loads(content)` yields a `str` when the model double-encodes the object.
2. `extract_response` → `arguments["response"]` is returned as-is when the model stringifies the nested object in the tool-call arguments.

`generate_llm_input()` is annotated `-> dict[str, Any]` but never checked. `runtime.py` then applied the value via `eval_item.model_copy(update={"inputs": ...})`, which skips pydantic validation, so the run only failed later at `MockingContext(inputs=...)` with:

```
1 validation error for MockingContext
inputs
  Input should be a valid dictionary [type=dict_type, input_value='{"messages": [{"role": ...}}', input_type=str]
```

That message reads like bad customer eval data. It is not: in prod (SRE-655743, 2026-09-07 UKS; a second org on 2026-09-08 NE) the **same eval item failed in one run and passed in the next** from the same eval set, and the failures landed 134 ms to 54 s after `EvalRun.Start`, before agent execution. It is the LLM's answer shape on that call.

## Fix

- **`_structured_output.py`**: `coerce_to_schema()` unwraps a `str` back to an object when the schema is `type: object` (handles single and double encoding). Scalar schemas pass through untouched, since the tool mocker legitimately returns scalars. Applied once at the `generate_structured_output` choke point so both the input mocker and the tool mocker benefit.
- **`_input_mocker.py`**: `_require_object()` enforces a dict result on both the fresh and the cached path and raises `UiPathInputMockingError` with an actionable message naming the returned type and a preview. A non-object is never written to the cache.
- **`runtime/runtime.py`**: `_generate_input_for_eval` re-validates the eval item through `model_validate` instead of `model_copy(update=...)`, and reports a `ValidationError` as `UiPathInputMockingError` so telemetry keeps one `ErrorType` for this class of failure.

## Tests

- `test_structured_output.py`: coercion of stringified and double-encoded objects, scalar schemas untouched, non-JSON strings returned unchanged, end-to-end through both the forced tool-call path (Claude) and the `response_format` path (OpenAI).
- `test_input_mocker.py`: non-object result raises with the new message and does not write to the cache; a stale non-object cache entry is rejected instead of leaking downstream.

`uv run pytest tests/cli/eval` (406 passed), `ruff check`, `ruff format --check`, `mypy` all clean. Version bumped to 2.14.15 with `uv lock`.

## Notes

- Once released, the prod URT low-code eval runtime still needs its `uipath` / `uipath-platform` pin bumped to pick this up (same gap tracked for SRE-656110).
- Jira: SRE-655743.


packages/uipath/src/uipath/eval/mocks/_structured_output.py
```diff
@@ -238,6 +238,30 @@ def _strategy_for_model(model: str | None) -> ToolCallStructuredOutput:
...
+    Handles single and double encoding. A string that is not valid JSON is
+    returned unchanged so the caller can report it.
+    """
+    if schema.get("type") != "object":
+        return result
+    for _ in range(2):
+        if not isinstance(result, str):
+            break
+        try:
+            result = json.loads(result)
```
AAgnihotry: ### P2 — F1: `coerce_to_schema` does not verify parsed JSON is actually a dict

When schema is `"type": "object"` and the model returns a stringified JSON array like `"[1, 2, 3]"`, `json.loads` succeeds and returns a `list`. The function returns it without checking it's a `dict`.

For `generate_llm_input`, `_require_object` downstream catches this. For `_llm_mocker.py` (tool mocking), a list could slip through when an object was expected.

**Risk:** Low — tool mocker passes inner response schemas typically not `"object"` at root, and input mocker has `_require_object`. But for correctness:

```python
try:
    parsed = json.loads(result)
    if isinstance(parsed, dict):
        result = parsed
    else:
        break  # wrong type, stop trying
except json.JSONDecodeError:
    break
```

---

### P2 — F2: Silent coercion without logging

When `coerce_to_schema` successfully unwraps a stringified object, it does so silently. If this happens frequently, it indicates a model/prompt quality issue operators should know about.

**Suggestion:** Add `logger.info` on successful coercion to make the behavior observable.

packages/uipath/src/uipath/eval/runtime/runtime.py
```diff
@@ -964,7 +965,18 @@ async def _generate_input_for_eval(
             expected_behavior=eval_item.expected_agent_behavior or "",
             expected_output=expected_output,
         )
-        updated_eval_item = eval_item.model_copy(update={"inputs": generated_input})
+        # model_copy(update=...) skips validation, so a malformed generated input
+        # would only surface later as an opaque MockingContext error. Re-validate
+        # through the model instead and report it as an input-mocking failure.
+        try:
+            updated_eval_item = type(eval_item).model_validate(
```
AAgnihotry, line 972: ### P2 — F3: No test coverage for this `model_validate` / `ValidationError` catch path

This is the last line of defense ensuring malformed input is caught before propagating deeper. A regression (e.g., reverting to `model_copy`) would silently reintroduce the original bug with no test to catch it.

**Suggestion:** Test `_generate_input_for_eval` with a mocked `generate_llm_input` returning an invalid value, asserting `UiPathInputMockingError` is raised with a descriptive message. Also a positive test confirming round-trip fidelity with aliased fields (if `by_alias=True` were accidentally dropped, every eval item with aliased fields would fail).

packages/uipath/src/uipath/eval/mocks/_input_mocker.py
```diff
@@ -23,6 +23,25 @@
 logger = logging.getLogger(__name__)
 
 
+def _require_object(result: Any) -> dict[str, Any]:
+    """Enforce that a simulated input is a JSON object.
+
+    ``generate_llm_input`` feeds ``EvaluationItem.inputs``, which must be a
+    dict. Without this guard a stringified object slips through
+    ``model_copy(update=...)`` unvalidated and only fails later as an opaque
+    ``MockingContext`` validation error (SRE-655743).
```
AAgnihotry: ### P3 — F4: Docstring references removed code path

This says "a stringified object slips through `model_copy(update=...)` unvalidated" — but this PR replaces `model_copy` with `model_validate` in `runtime.py`. The docstring describes the historical bug, not what would happen today if `_require_object` were removed (the `model_validate` call would catch it, but with a less actionable error).

**Suggestion:** Rewrite to describe the current architecture: "Without this guard, a non-dict value would be caught by `model_validate` in the eval runtime, but the resulting `ValidationError` would lack context about *why* the input was malformed. This guard produces an actionable error message identifying the mocking step as the source."

packages/uipath/src/uipath/eval/mocks/_structured_output.py
```diff
@@ -238,6 +238,30 @@ def _strategy_for_model(model: str | None) -> ToolCallStructuredOutput:
...
 
 
+def coerce_to_schema(result: Any, schema: dict[str, Any]) -> Any:
+    """Unwrap a JSON object the model returned as a string.
+
+    Models sometimes stringify a nested object inside tool-call arguments, or
+    double-encode ``response_format`` content, so the caller receives a ``str``
+    holding JSON instead of the object it asked for. Only applies when the
+    schema asks for an object; scalar and array schemas pass through untouched
+    because a string is a legitimate value for them.
```
AAgnihotry: ### P3 — F5: Docstring imprecision for array schemas

"scalar and array schemas pass through untouched because a string is a legitimate value for them" — a string is not a legitimate value for an array schema (JSON arrays and strings are distinct types). The real reason is the observed bug only affects object schemas.

**Suggestion:** "Only applies when the schema asks for an object, which is the type affected by the stringification bug. Other schema types are left untouched to avoid unintended coercion."

packages/uipath/src/uipath/eval/mocks/_input_mocker.py
```diff
@@ -132,15 +151,17 @@ async def generate_llm_input(
...
-                return cached_response
-
-        result = await generate_structured_output(
-            llm,
-            [{"role": "user", "content": prompt}],
-            schema=input_schema,
-            response_format_name="agent_input",
-            description="Return the simulated agent input matching the required schema.",
-            completion_kwargs=completion_kwargs,
+                return _require_object(cached_response)
```
AAgnihotry: ### P3 — F6: Stale cache entry rejected but not evicted

When `_require_object` rejects a stale cached string, the bad entry remains in cache. Every subsequent call with that key hits cache, gets rejected, and errors. Low priority since cache is ephemeral (per eval set run), but an eviction or error message mentioning cache would improve UX.

**Suggestion:** Either evict the bad entry before raising, or include "this value was loaded from cache" in the error message.

AAgnihotry (commented):
## Comprehensive PR Review — 6 specialized passes

**Verdict: Approve** — well-scoped bug fix with solid defense-in-depth. No blocking issues.

| Severity | Count |
|----------|-------|
| P0 (Critical) | 0 |
| P1 (High) | 0 |
| P2 (Medium) | 3 |
| P3 (Low) | 3 |

### Slop Meter

🟢 **Clean** — Slop: 2/100 · Tunnel-vision: 0/100 (only version literals flagged)

### Findings Index

|---|-----|-----------|----------|
| F1 | P2 | `_structured_output.py:259` | `coerce_to_schema` does not verify parsed JSON is a dict |
| F2 | P2 | `_structured_output.py:259` | Silent coercion without logging |
| F3 | P2 | `runtime.py:972` | No test coverage for `model_validate` / `ValidationError` catch |
| F4 | P3 | `_input_mocker.py:29` | Docstring references removed `model_copy` path |
| F5 | P3 | `_structured_output.py:248` | Docstring claims string is legitimate for array schemas |
| F6 | P3 | `_input_mocker.py:154` | Stale cache entry rejected but not evicted |

All finding details are inline below.

### Strengths

- **Defense-in-depth:** Three layers — coercion at structured output, type enforcement at input mocker, validation at runtime. Even if one fails, the next catches it.
- **Actionable error messages:** `_require_object` includes actual type + truncated preview. Clear signal vs the prior opaque `MockingContext` error.
- **Cache poisoning prevention:** Bad values never written to cache.
- **Broader benefit:** `coerce_to_schema` at the `generate_structured_output` choke point means `_llm_mocker.py` also benefits.
- **Correct Pydantic alias handling:** `model_dump(by_alias=True)` + `model_validate` with `populate_by_name=True` round-trips correctly.

### Recommended follow-ups (not blocking)

1. Add `isinstance(parsed, dict)` check after `json.loads` in `coerce_to_schema` (F1)
2. Add `logger.info` when coercion fires (F2)
3. Add a test for `_generate_input_for_eval` `model_validate` path in `runtime.py` (F3)

🤖 Review generated with Claude Code using 6 specialized agents (code-reviewer, test-analyzer, silent-failure-hunter, comment-analyzer, deep-review, adversarial-review)

packages/uipath/src/uipath/eval/mocks/_structured_output.py
mjnovice: Done in 5f1694cc. F1: `coerce_to_schema` now only returns the parsed value when it is a dict; stringified arrays/scalars (single or double encoded) are returned unchanged so the caller reports the real value. F2: added a `logger.info` on every unwrap, including the encoding depth. Tests added for both.

packages/uipath/src/uipath/eval/runtime/runtime.py
mjnovice: Done in 5f1694cc. Added `tests/cli/eval/test_eval_runtime_input_mocking.py` which drives `_generate_input_for_eval` on a real `UiPathEvalRuntime` with `generate_llm_input` patched: one test asserts the full aliased-field round trip (would fail if `by_alias=True` were dropped), and a parametrized test covers a non-string key, a list and a stringified object, asserting `UiPathInputMockingError` with the schema message.

packages/uipath/src/uipath/eval/mocks/_input_mocker.py
mjnovice: Done in 5f1694cc. Rewrote the docstring along the lines you suggested: it now describes the current flow (runtime `model_validate` would catch it, but without saying why) and why the guard fails early with an actionable message.

packages/uipath/src/uipath/eval/mocks/_structured_output.py
mjnovice: Done in 5f1694cc. Rewrote the docstring per your suggestion; it no longer claims a string is legitimate for array schemas.

packages/uipath/src/uipath/eval/mocks/_input_mocker.py
mjnovice: Done in 5f1694cc. Went with self-healing rather than raise: the cached path now runs `coerce_to_schema` too, so a stale stringified object is unwrapped and re-cached. A cached value that cannot become an object is logged at warning level, ignored, and regenerated, with the fresh result overwriting the stale entry. Tests cover both paths.

packages/uipath/src/uipath/eval/runtime/runtime.py
```diff
@@ -964,7 +965,18 @@ async def _generate_input_for_eval(
...
+                {**eval_item.model_dump(by_alias=True), "inputs": generated_input}
+            )
+        except ValidationError as e:
+            raise UiPathInputMockingError(
+                "Simulated input does not match the evaluation item schema: "
+                f"{format_exception_message(e)}"
+            ) from e
```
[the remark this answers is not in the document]
mjnovice: Addressed in 5f1694cc together with the F3 thread below: `tests/cli/eval/test_eval_runtime_input_mocking.py` now exercises `_generate_input_for_eval` with a non-string key, a list and a stringified object, asserting `UiPathInputMockingError` is raised.

AAgnihotry (approved):

